PR #26, opened 2017-10-24
modify folder hierarchy

millerthegorilla:
Changed code to allow many sub-folders to exist.
Added some comments about why etc, ignore them if you want.  I'm new to java and android as of yesterday, so please take a good look and test thoroughly.


millerthegorilla:
I'd also like to examine the way in which song times for the autoscroll are saved and how they change when the screen orientation changes, as I keep losing my song times.  I'd like to suggest an increase in the size of the timer font, and perhaps make a fragment with a slider to adjust the speed.

thebigg73:
I've updated the autoscroll values to hopefully make the saves more often.  It should be easy enought to incorporate a save on an orientation change.

I'm getting ready for a release of v4.0.9, so I haven't pulled your code yet.  Really impressed with the work you've done though - especially if you are new to Java/Android!  I've been teaching myself Android for the last few years using this app as a learning platform.  As you've probably seen, my code probably isn't the cleanest, or best, so feel free to improve it.  Once the update is live (and working), I pull your changes into my live testing branch to see how it looks before hopefully merging with the master.
Have you tested any of these changes on your device yet?

millerthegorilla:
I haven't tested on my device yet, not certain how to do that.  I've just been reading about performance on android devices, which is going to be more of an issue than I am used to.  I don't do a lot of developing, but I've been doing it for a long, long time, and although this is my first time on Java, I've programmed in c++ a bit before (I studied it carefully once upon a time) and I have done a little intensive javascript/actionscript/ruby/python etc.
I've just tidied up the code and removed some of the more spurious comments and put some more informative ones.  I'll have a go at a performance benchmark on the stream vs collections tomorrow, and do a bit more reading about variable storage/caching etc.  I'll also learn to test on my device as well.

millerthegorilla:
Glad you like it btw.  I'm also self-taught, but I happened to be doing an unrelated degree when I began, so I learnt quite formally.  I haven't worked in the field, but have maintained my interest.  I'd love to have a go at refactoring the code, so when I've got some spare time, I'll start taking a look.

PR #43, opened 2018-06-22
#37: close DrawerLayout's on back button if open

### LINKED ISSUE #37: Usability: behaviour of back button
OpenSongTablet has some kind of screen overlay popups.
If you click the "note" symbol from the toolbar you get this overlay popup:
screenshot_20180620-135630
If you click the menu button from the toolbar you get this overlay popup
screenshot_20180620-140634

Now my expectation as a user would be that the back button would close these overlay popup. However it instead asks if I want to close the app:
screenshot_20180620-135804

hohwille:
I have fixed #37 both for Presenter and Stage mode. I do not like the redundancy but I wanted a minimal change. Extracting a super-class to share common code between the two modes would be a different story/issue. I tested and debugged my change and it works as expected to me.

### DIFF
--- a/app/src/main/java/com/garethevans/church/opensongtablet/PresenterMode.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/PresenterMode.java
@@ -3783,6 +3783,17 @@ public class PresenterMode extends AppCompatActivity implements MenuHandlers.MyI
             mp.reset();
             mpTitle = "";
         }
+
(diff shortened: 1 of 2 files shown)


thebigg73:
I haven't forgotten!  I'm on holiday just now and away from my development laptop for a couple of weeks.  I'll merge this once I've tested myself, but it sounds like a straightforward merge.

thebigg73 (approved):
Straightforward and sensible

PR #48, opened 2018-07-29
show capoinfo always in toolbar like key

hohwille:
When I go to the chords menu, I can either choose to show capo infos what enforces to show the transposed chords in red. I can choose to also show the original chords in yellow in addition but there is no option to show only the original chords from the chordpro/opensong text together with the capoinfo.
Therefore I added the capoinfo to the toolbar just like the key is currently displayed.

In case you do not like it, you might consider to always show the capoinfo on the right on top of the quick launch buttons no matter what option is configured in the chords menu. IMHO if the capo info is set it should always be present somewhere so I cannot miss it.


hohwille:
In order to play around a little and get more into android development I also created a feature branch in my fork that adds a new option which also allows to switch the order of native and capo chords and then showCapoAndNativeChords option only decides if the second of these chord lines is actually added and I then missused the showCapoChords option to toggle the capo visibility:
https://github.com/hohwille/OpenSongTablet/tree/capo-show-option

However, I would rather question if it is better to simply always show the capoInfo if capo is set in the song (e.g. instead of this PR or in addition). There is actually a flag `FullscreenActivity.showCapo` but for some reason I could not find a way to toggle this without activating the capo chords as a user (not as a developer). As a guitar player I always want to have my songs with chords that are easy to play and therefore I note them as native chords in the song. But to play the song in the proper key and in a way to properly sing along with it, I then note the fret where I have to put the capo. Therefore the most regular use-case for me is to still see the native chords and also get the capo info displayed to not forget putting the capo in the right place. Having the option to also show the capo chords in the transposed key is a very nice feature but not my common use-case.

I do not intend to create a pull request from the feature branch linked above... After digging through all the code I got the impression that OpenSongTablet is already dying of too many features with options to tweak each and everything. Flexibility is nice but simplicity (KISS) and focus on the main features also has its own value. So IMHO having the ability to customize the font sizes, individual theme colors or even the percentage of the menu width is something that 99% of the users do not need.

PR #57, opened 2019-02-04
V5 prep

thebigg73:


thebigg73:
Updated master with V5 prep (in beta testing)

PR #61, opened 2019-06-12
Fix extraction of superscript characters from SongSelect

ermshiperete:
- recently when importing songs from SongSelect the chords were messed
  up because of sprinkled sup tags in the text. This change removes
  the superfluous tags.
- also add some unit tests for the chord extraction.

---
This change is [Reviewable](https://reviewable.io/reviews/thebigg73/opensongtablet/61)


thebigg73:
I've added these changes to the master branch.  I've added a few extra bits as well!
Many thanks.

PR #64, opened 2020-07-23
master

iv-gha:
Hello Gareth,
Feature behaviour improvements and some new things for consideration.
Please note there are comments starting 'IV -' that are intended to inform and not be retained if merged.
Hopefully I am getting this github thing right!
Kind regard
Ian V


thebigg73:
Hi Ian,

Believe it or not, this is quite new to me as well!  Not many users have
taken the plunge to help with coding.  I'll have a look today and get back
to you

Many thanks for getting involved!

Gareth

app/src/main/java/com/garethevans/church/opensongtablet/Metronome.java
```diff
@@ -225,34 +227,25 @@ static void startstopMetronome(Activity activity, Context c, boolean showvisual,
...
             try {
                 metroTask.executeOnExecutor(AsyncTask.THREAD_POOL_EXECUTOR);
             } catch (Exception e) {
                 Log.d("d","Error starting the metronome");
             }
             startstopVisualMetronome(showvisual,metronomeColor);
-
-        } else if (checkMetronomeValid(c) && StaticVariables.metronomeonoff.equals("on")) {
+        // IV - A stop perhaps does not need to consider if it is valid
+        } else if (StaticVariables.metronomeonoff.equals("on")) {
```
thebigg73, line 240: Going to need to look at this further.  Not sure if it is this change or another, but now when a song is playing a metronome, swiping to the next song does not display it and the metronome keeps playing.
Original action was to display a warning that it wasn't set, but the next song did load

thebigg73: Happy not to open the popup (just display toast that metronome is not set), then stop the metronome (currently in my logic it wrongly keeps playing)

app/src/main/java/com/garethevans/church/opensongtablet/OptionMenuListeners.java
```diff
@@ -2629,9 +2630,13 @@ public void onClick(View v) {
         autoscrollActivatedSwitch.setOnCheckedChangeListener(new CompoundButton.OnCheckedChangeListener() {
             @Override
             public void onCheckedChanged(CompoundButton compoundButton, boolean b) {
-                StaticVariables.clickedOnAutoScrollStart = b;
-                if (!b && mListener!=null) {
-                    mListener.stopAutoScroll();
+                // gesture contains the start/stop logic.
+                if (mListener != null) {
+                    // Request a start/stop to get to the desired state
+                    if ((!StaticVariables.isautoscrolling && b) || (StaticVariables.isautoscrolling && !b)) {
```
thebigg73, line 2636: Maybe change to
if (StaticVariables.isautoscrolling!=b) {

app/src/main/java/com/garethevans/church/opensongtablet/PopUpChordsFragment.java
```diff
@@ -84,7 +85,14 @@ public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
...
         TextView title = V.findViewById(R.id.dialogtitle);
-        title.setText(Objects.requireNonNull(getActivity()).getResources().getString(R.string.chords));
+
+        // Title changed to reflect Native or Capo chord display
+        if (StaticVariables.showCapoInChordsFragment) {
+            title.setText(Objects.requireNonNull(getActivity()).getResources().getString(R.string.showcapo));
+        } else {
+            title.setText(Objects.requireNonNull(getActivity()).getResources().getString(R.string.showchords));
+        }
+
```
thebigg73, line 95: I really like this suggestion, however, it doesn't seem to do anything yet?  Need to figure out where/how StaticVariables.showCapoInChordsFragment is set

app/src/main/java/com/garethevans/church/opensongtablet/PopUpPageButtonsFragment.java
```diff
@@ -150,7 +151,8 @@ public void onClick(View view) {
         custom3Visible_Switch.setText(c3);
         custom4Visible_Switch.setText(c4);
 
-        enableordisablegrouping();
+        // IV - collapsing and expanding groups means a block is no longer needed
+        //enableordisablegrouping();
```
thebigg73, line 155: This was to stop grouping of information and custom buttons from being active if all buttons are grouped in one setting as changing them would have no effect on what is seen if all buttons are already grouped

Further review required

app/src/main/java/com/garethevans/church/opensongtablet/ProcessSong.java
```diff
@@ -1193,6 +1193,8 @@ private TableRow capolinetoTableRow(Context c, Preferences preferences, int lyri
             capobit.setTextSize(fontsize * preferences.getMyPreferenceFloat(c,"scaleChords",1.0f));
             capobit.setTextColor(lyricsCapoColor);
             capobit.setTypeface(StaticVariables.typefaceChords);
+            // IV - Choice to make bold
+            capobit.setPaintFlags(capobit.getPaintFlags()| Paint.FAKE_BOLD_TEXT_FLAG);
```
thebigg73, line 1197: Need to check where the option to switch this on and off is (same for headings)

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
```diff
@@ -1825,26 +1890,32 @@ private void setupPageButtons() {
...
         setButton.setOnClickListener(view -> {
             CustomAnimations.animateFAB(setButton,StageMode.this);
             FullscreenActivity.whattodo = "editset";
             openFragment();
         });
+        // IV - New Long click to set item variation
+        setButton.setOnLongClickListener(view -> {
+            FullscreenActivity.whattodo = "setitemvariation";
+            openFragment();
+            return true;
```
thebigg73, line 1917: Like this option.  However, this also fires the page gesture (long pressing on a song on the page adds it to the set).
This brings up the add variation, but also adds new song to set.  Because the set popup shows first, the new song added doesn't show.

thebigg73: Actually ignore that!  This happens with all the page buttons - it is possible to long click them and the page together with a fat finger!  Might have to increase the padding around the buttons to capture button presses on their own.

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
```diff
@@ -1921,8 +1990,18 @@ private void setupPageButtons() {
...
+        extraButton.setOnLongClickListener(view -> {
+            CustomAnimations.animateFAB(extraButton,StageMode.this);
+            preferences.setMyPreferenceBoolean(StageMode.this, "pageButtonGroupMain", !(preferences.getMyPreferenceBoolean(StageMode.this, "pageButtonGroupMain", false)));
+            // Need to redo OnClick button action
+            setupPageButtons();
+            groupPageButtons();
+            // Make sure all dynamic (scroll and set) buttons display
+            onScrollAction();
+            return true;
+        });
```
thebigg73, line 2003: Yes, I like this and see what you did on the PopUpPageButtonsFragment that I queried.

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
```diff
@@ -2599,6 +2772,91 @@ private void customButtonAction(String s) {
...
+                    openFragment();
+                }
+                break;
+
+            case "showchords":
+                // Display the native chords
+                StaticVariables.showCapoInChordsFragment = false;
+                FullscreenActivity.whattodo = "page_chords";
+                openFragment();
+                break;
```
thebigg73, line 2830: This explains the PopUpChordsFragment behaviour.  Makes sense now.

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
```diff
@@ -7146,24 +7233,42 @@ protected void onPostExecute(String s) {
...
                     }
 
-                    // Automatically stop the metronome if it is still on
-                    if (StaticVariables.metronomeonoff.equals("on")) {
-                        // Stop it
-                        Metronome.startstopMetronome(StageMode.this, StageMode.this,
-                                preferences.getMyPreferenceBoolean(StageMode.this,"metronomeShowVisual",false),
-                                defmetronomecolor, preferences.getMyPreferenceString(StageMode.this,"metronomePan","C"),
-                                preferences.getMyPreferenceFloat(StageMode.this,"metronomeVol",0.5f),
-                                preferences.getMyPreferenceInt(StageMode.this,"metronomeLength",0));
```
thebigg73, line 7156: I wonder if this is the cause of the metronome issue on moving between songs?

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
```diff
@@ -7467,44 +7566,138 @@ protected void onPostExecute(String s) {
...
+                ShowToast.showToast(StageMode.this);
+                StaticVariables.pedalPreviousAndNextNeedsConfirm = false;
+                pedalPreviousAndNextNeedsConfirmHandler.postDelayed(pedalPreviousAndNextNeedsConfirmRunnable, 10000);
+                StaticVariables.pedalPreviousAndNextIgnore = true;
+                // Use a runnable to end the ignore period.
+                pedalPreviousAndNextIgnoreHandler.postDelayed(pedalPreviousAndNextIgnoreRunnable, 2000);
+            } else {
+                try {
+                    // If pedal used again in the ignore period - extend the ignore period.
+                    pedalPreviousAndNextIgnoreHandler.removeCallbacks(pedalPreviousAndNextIgnoreRunnable);
```
thebigg73, line 7587: Not seen how or why this is implemented yet

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
iv-gha: I had thought I had changed to gesture use for all metronome calls - have I missed off a change!  Let me look.

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
iv-gha: You will :-) It is in pedal press logic.  You will need a pedal to see it in action.  It perhaps needs to be an option that can be used.  It addresses the panic cause by a wrong foot press changing songs during performance.

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
iv-gha: Strange, this will happen for any long press button action?  I added an ignore section around button use to try to avoid main screen mis-fires!

I think main screen long press gesture is a big problem - it fires when not wanted a lot.  Fingers are not precise!  It may interfere with dragging in autoscroll etc.  My vote is to remove it (I tend to turn it off when using live) as it fires a little too often when using buttons etc.  Particularly as we like button long press actions!

app/src/main/java/com/garethevans/church/opensongtablet/ProcessSong.java
iv-gha: There are no options yet.  This change presents songs more like a typical song print, where chords and section titles are bold.  This and use of Roboto font (much better width to the space character) gives an on screen display close to a typical printed page look (probably a users initial expectation).

Yes these should be settings - particularly as current users might not like!  Did not want to fiddle with menus!

app/src/main/java/com/garethevans/church/opensongtablet/PopUpPageButtonsFragment.java
iv-gha: This is consistent with the 'new' ability to expand/collapse groups.

app/src/main/java/com/garethevans/church/opensongtablet/PopUpChordsFragment.java
iv-gha: It is in the long press for 'show capo chords' button.

The menu may need an option to display capo chords.  The separate 'chords' button could be retired as not necessary. That is menu stuff which I did not touch.

app/src/main/java/com/garethevans/church/opensongtablet/OptionMenuListeners.java
iv-gha: Sure, I kinda coded my thoughts a little too much!  I think there is another case of of this 'over thinking'.

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
iv-gha: I have changed over to gesture7 use...  I  do not see this stop.  So would need to understand the metronome settings you are using for the original and new song to try to replicate.  What are the settings?
Are you testing against a full merge?

app/src/main/java/com/garethevans/church/opensongtablet/OptionMenuListeners.java
thebigg73: Glad I'm not the only one!

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
iv-gha: OK

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
iv-gha: OK

app/src/main/java/com/garethevans/church/opensongtablet/PopUpChordsFragment.java
thebigg73: Got it.  I think leave it as it is for now.

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
thebigg73: I personally like the long press gesture.  However, might be sensible to set the default behaviour for new users to off?

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
thebigg73: Hi Ian,
Yes, I created a new pull request from your fort to a branch named IanTest.  I accepted this extra pull request so I could test without affecting the master branch just now.
Have a song with metronome values set.  Start the metronome (either using the popup start button or long pressing).
Then try to swipe to the next song.  Two issues:
1.  If you are in a set, nothing appears to happen when you swipe.  As soon as you stop the metronome, the swipe is called.
2.  If you are not in a set, the song is swiped away, but the new one isn't loaded (black screen) until you stop the metronome.

Additionally, the metronome should be stopped if it hasn't been set for the newly loaded song after swiping (although this bit was also an issue in my code).

I think this is the only issue to resolve - really well done on figuring out my ugly, messy code and making it better!

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
thebigg73: Still can't see this happening (tried it with a pedal) - I'm never shown the 'Are you sure' dialog.
Once we figure this out, I think that it should be a selectable option.  Most users would probably not want that alert showing when first running the app, well until they get their first issue of false presses on stage (my keyboard player does that to me every so often!)

iv-gha: Must be pedal.  Do not use the pedal for 10+ seconds (like when performing a song).  Do a previous or next pedal action.  No go?  Strange.

Selectable yes - just a wrap with a preference and more menu stuff!

I was also not certain on the 2s delay. But if you are pedal press happy it needs the intentional break.

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
iv-gha: Mess is much better than nothing!  I can work with moderate mess.  Just a few tweaks to assist.

The code stops a playing old song metronome using gesture7 and then, if needed, gesture7 to start the new songs metronome.  So problem is in gesture7 and supporting code.

                    // Do not touch on a reload
                    if (!StaticVariables.reloadOfSong) {
                        // Stop it
                        // IV - gesture used for stop and start
                        if (StaticVariables.clickedOnMetronomeStart) {
                            gesture7();
                            // Metronome Was playing - if requested autostart start the metronome for a new song
                            if (preferences.getMyPreferenceBoolean(StageMode.this, "metronomeAutoStart", false) &&
                                    FullscreenActivity.isSong) {
                                // Start it
                                gesture7();
                            }
                        }
                    }

Perhaps you can debug to see what is going on?  Perhaps StaticVariables.clickedOnMetronomeStart is not being set correctly elsewhere.

I do the test.  In a set and move to a song with metronome OK.  Move to song with not set metronome gives toast.   What are your detailed metronome settings please.

thebigg73: That's weird that it is working in yours.  I'm going to try a reset/invalidate of the caches in Android Studio as the code is showing me some errors, yet still running - something isn't right at my end!

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
iv-gha: It needs to not mis-fire.  I have had it fire during testing of dragging autoscroll.  The gesture could be turned off in autoscroll as dragging can lead to accidental long press.  Must though only fire if well away from buttons or top bar icon / features.   Not sure how you get there :-)  Perhaps restrict it to the area below top bar, below sticky x cancel, above bottom buttons and to right of other buttons. Fun.

app/src/main/java/com/garethevans/church/opensongtablet/Metronome.java
iv-gha: OK

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
iv-gha: OK

iv-gha: Hang on... I seem to remember that during testing the emulator struggles with the metronome code.  Try a build and run on a device.  Also try your original code on the emulator to see (hopefully!) if it also mis-behaves.

iv-gha: Just tried here.  Yes, the emulator does not do well!  Slow response and the occasional crash and only moves song when metronome stopped.  No problems on Fire HD 10 or Moto G6 Play.   Did not investigate the cause.

iv-gha: I was reporting it working for my Fire HD 10.

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
thebigg73: Ok I've figured this out - tracing back through your code (something I realise takes time!), this only happens when performing songs in a set.  This might be useful for non set work as well, e.g. if you go off the planned set and move to another song.
I'll amend my IanTest branch and after pulling your changes to Master, I'll pull request my IanTest to master as well.
I'll add a toggle setting too

iv-gha: Ahh.  I had forgotten.  Yes it is currently for set use only.  I can see your point that it might be useful elsewhere but (I think) I thought it was best not to intrude too much.   Perhaps an option to protect all song changes or just when in a set. Solved.

A quirk that I noticed is that you can wander into a set song whilst moving around the song list and the set logic then takes over!  So the next move is within the set.

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
thebigg73: Changed the default behaviour to off in my branch.  Will push changes once pull request completed

app/src/main/java/com/garethevans/church/opensongtablet/ProcessSong.java
thebigg73: Added a toggle (in fonts settings page) to switch this on.  Added to IanTest branch and will push to master after completing pull request.

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
thebigg73: I program on Linux and can't use the emulator!  My test device is a Samsung Galaxy Nexus 10, which despite being old (and updated to Marshmallow using custom ROM), works really well.  Better performance than my newer Samsung Galaxy Tab 10.  I'll continue to debug this jut now.

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
thebigg73: Believe it or not, that was intentional!  Often I deviate from a set ny picking another song from the song menu (setView is then off).  Then I pull up a song from the song menu that is back in the set, then the app puts me back into set mode.
I did this because when not moving through a set, navigating back to a set item always took me ages to find the where the song was.  It's quicker to find it in the song menu and have the fact that it is labelled as a set item throw me back to the correct point of the set.
If you pick a song that is in the set from the menu, it should also reset your position in the set to the current item (or first version of it if you've put it in multiple times).

The fix to add the are you sure prompt was pretty straightforward, so I've added.
Once I merge your pull request to master and push my additions, I'll get you to update your fork and test the changes.

Just the metronome issue to fix and we're there!

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
iv-gha: OK.  There has been a slight reorg of code so the stop code is called by gesture 7 - but the call appears to be no different.  Strange.

iv-gha: Thanks for letting me know your dev env.  FYI.  I am Windows 10 / AS 4.0.1 with devices emulator, Fire 10 HD, Fire 8  and Moto 6 Play.  This is my first coding using AS and Android/Java.  So thanks for giving me a good reason to get into a new language.

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
iv-gha: I think this might be the the most subtly complex bit of code in the commit - with runnables being fired and reset to get a reasonable effect.  It is aimed at gatekeeping the first song move but then allowing the user to move around freely - as they might then want to quickly move around in a set using the pedal.  It becomes active again 10s after they stop moving.  So not every page move but hopefully the most 'dangerous' one!   The intent is as a 'stage' performance protection feature.

iv-gha: As a further comment - the absence of previous and next buttons when not in a set could be considered.  Maybe always have buttons but somehow indicate use in a set (a different icon?).  A thought.

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
thebigg73: Done some testing.  If I stop the metronome just before starting the loadsong_async task the song loads fine, but the metronome isn't resuming on the next song.  If I leave the start stop logic to be run as part of the async task itself, the async task never starts until I stop the metronome.  Very strange - I wonder if there is a limit to the number of tasks that can run simultaneously on some devices (since you don't see this issue).
I'm heading out just now, but at least I've figured out a potential fix - split the following bit up into a stop (before load) and  restart if required after load:

if (!StaticVariables.reloadOfSong) {
                            // Stop it
                            // IV - gesture used for stop and start
                            if (StaticVariables.clickedOnMetronomeStart) {
                                gesture7();
                                // Metronome Was playing - if requested autostart start the metronome for a new song
                                if (preferences.getMyPreferenceBoolean(StageMode.this, "metronomeAutoStart", false) &&
                                        FullscreenActivity.isSong) {
                                    // Start it
                                    gesture7();
                                }
                            }
                        }

iv-gha: Sounds like a way forward.  Great.

iv-gha: Am I right in thinking metronome is not called async like other actions?  Perhaps it should be to stop any interference / delay to song display?

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
iv-gha: image
A suggestion for an icon!

iv-gha: image
image

Perhaps simpler is better.

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
iv-gha: My feeling is to async metronome and leave the call as-is. So do what you have done with the other actions.  This will keep the logic very clear (no splitting of activity) and get out of the way of song load (which is good).  This has the advantage of keeping the approach consistent across actions. Kind regards. Ian.

thebigg73: Yeah, the metronome is already Async - initiated and run from the Metronome.java class.  It is actually split into two async tasks - one for the metronome, which doesn't need a UI update, so all done in doInBackground task and one that handles the visual metronome, which flashes the actionbar, so does the logic in a doInBackground task and the UI update in OnProgressUpdate,.  They used to be merged (using if... statements to decide which bits to fire), but I definitely felt performance was more solid (less timing issues) when split up and called individually.

thebigg73: Ok, I've tracked my way through the app using logcat statements and found that the app wouldn't get past the animateInSong without the metronome being stopped first.  Not quite sure why!  So, I've moved the restart gesture7 logic to after the animateIn (which makes sense as displaying the song should be the highest priority in the chain).  It all works now!!!

I'm just going to make a few other changes (to address Android 8+ devices showing blank storage locations when using the file picker library), then I'll merge the pull requests to the master and get you to update for testing at your end.

iv-gha: Ahh.  Should have looked harder!  Leave it with you.  Have fun!

iv-gha: Great stuff. 

Just a note for testing... i have focused on songs so not PDF or image which I know are supported.  For me they are not good song types so I do not use but please do make sure to regression test (there is lots of things I don't do).

Thanks for being so receptive to the new features of the commit.

thebigg73 (approved):
Pull request accepted.  Some additional code will be pulled following this to address some issues.
Many thanks for your contribution!!

PR #69, opened 2020-07-28
Page change protect / fixes

thebigg73:
GE Pull request to test branch before master

Fixes:
1. Adjustments page change protection to apply across folders/sets and extended to all song types with fix of missing set buttons if feature not in use.
2. Song export
- corrections to typos for preference names in popup (options now save)
- txt export layout corrections for section breaks and chords alignment
- Resolved conflict PDF and png process preventing correct export
3. presentation mode option menu returns to 'MAIN' after use (same as recent Stage mode change).

Enhancements:
1. Autoscroll will pause on any pedal use.  A user can configure a long press scroll up to go to the top of the song and keep it pressed to hold the song paused at the top.
On releasing the pedal autoscroll will start.  This gives manual control over when autoscroll starts.
2. A lyrics only song display will compress white space and remove word splits of format " - " (at least one space on either side.  An attempt to further beautify.

### DIFF
--- a/app/src/main/java/com/garethevans/church/opensongtablet/ExportPreparer.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/ExportPreparer.java
@@ -183,7 +183,7 @@ class ExportPreparer {
         Uri onsong = null;
         Uri image = null;
         Uri pdf = null;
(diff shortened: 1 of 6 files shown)


thebigg73 (commented):
Bring in to test
Initial review looks ok though

PR #71, opened 2020-07-29
Fixes

iv-gha:
1. Short key press seen when holding a pedal long press and using a fragment fixed.
   Fix: On fragment entry short key presses are temporarily ignored, the false key press is ignored in this period.
2. Rapid song changes can result in incorrect first/last song toast messages.
   Fix:  Code is confused by a song being partially loaded in the background.  Relevant next and previous code now runs only if not in the middle of a song load.
   Negatives: One or more changes in a rapid sequence of changes will be ignored
3. Corrections to chord and lyric positions for chordpro conversions.
Ian V

### DIFF
--- a/app/src/main/java/com/garethevans/church/opensongtablet/ChordProConvert.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/ChordProConvert.java
@@ -308,9 +308,10 @@ class ChordProConvert {
     String extractChordLines(String s) {
         StringBuilder tempchordline = new StringBuilder();
         if (!s.startsWith("#") && !s.startsWith(";")) {
+            // IV - Add a leading space - the effect is to fix a chord mis-alignment
+            s = " " + s;
             // Look for [ and ] signifying a chord
             while (s.contains("[") && s.contains("]")) {
-
                 // Find chord start and end pos
                 int chordstart = s.indexOf("[");
                 int chordend = s.indexOf("]");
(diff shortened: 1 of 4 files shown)


thebigg73 (commented):
All look good - I'll test, but safe to merge with the master branch.
Well done on fixing the conversion between OpenSong and ChoPro - that was irritating me!

iv-gha:
Hello Gareth,

Great.  The tweaks looked reasonably safe to me - I am glad you agreed.

I think we have made some good improvements which should help users appreciate your hard work more!

I get the impression there are two places of chordpro conversion? I think I have improved one.

Kind regards
Ian V

iv-gha:
Apologies,

Spotted a mistake...

        // IV - Block false short key press if fragment used during long press
        StaticVariables.blockActionOnKeyUp = true;
        Handler resetBlockActionOnKeyUp = new Handler();
        resetBlockActionOnKeyUp.postDelayed(() -> {
            StaticVariables.blockActionOnKeyUp = false;
            prepareOptionMenu();
        }, 300);

Please remove the line  prepareOptionMenu();  when you are next in - I copied a handle block from elsewhere and left it in?  Not dangerous but not necessary.

Kind regards
Ian V

thebigg73:
Yeah - I'm pleased with the improvements we've made!  The app has an OnSong (iOS software) conversion feature.  This is pretty much the same file format as ChordPro but with a couple of extras.  If I remember rightly it sends the conversion to the chordpro convert after doing a couple of fixes first.
With the development of features over many years the code has become messy and there are bits of code duplication all over the place.  That's partly what I'm trying to fix in the MaterialApp branch, along with making it more pleasing to the eye and more user friendly.  Currently it is a bit 'death by menus'!

I'll remove that extra line as well.

PR #73, opened 2020-08-03
Test IV changes

thebigg73:
Just checking potential issues


iv-gha:
Thanks Gareth.  I had trouble doing the pull (needed to cancel one and sync closer to your master) so please makes sure you use the latest.

I have seen some changes for PDF pad playing - hopefully you have it working as needed.

Kind regards
Ian

thebigg73 (commented):
Showed conflicts when initialising the pull request, however, on checking, nothing particularly wrong that I can see looking at the coding.

thebigg73:
Found one conflict in Stage Mode.  I think I've fixed it and it reported that it would update your master branch with the fix.  It was around line 5422 in Stage Mode
resetBlockActionOnKeyUp.postDelayed(() -> {
            blockActionOnKeyUp = false;
        }, 300);
The conflict had the older StaticVariables.blockActionOnKeyUp=false;.  I removed the StaticVariables reference as I'm sure you moved this to a private boolean in StageMode.  If not, I/we'll need to change it back.

thebigg73:
Hi Ian,

The issue might have been my fault - When I made my last changes on my
local system, I hadn't pulled in some of the files we'd previously changed,
so when I tried to push my updates to the master it threw an error and
showed conflicts.  I had to then merge them and check everything manually -
note to self: always check my local and remote (origin) are fully
synchronised before pushing updates!
I think these are the differences that have triggered apparent 'extra'
files in your earlier pull request.  Unfortunately it also meant that your
last changes didn't appear in V5.1.9 which is currently in beta - sorry!
 However, your earlier fixes and my changes did make it and I'm pleased to
say it hasn't seen any crashes or received any negative feedback as yet, so
I'm likely to move it (with these recent changes) to the stable channel on
Google Play.
Just before you submitted the second pull request, I created a pull request
from your code to my test branch, fixed one conflict (with my branch, made
a comment - only affected me, not you).
I'm really impressed with the OpenSong->ChordPro and ChordPro->OpenSong
conversion work - it is now seamless!
Chord positions work well in this version - well done, however, I found a
bug in the chord identification logic that throws an error (and displays no
song) if a line is completely empty (substring out of bounds) - in
ProcessSong.java  Fixed this with a check.
Also adjusted code when comparing a string (bits != "" should be
!bits.isEmpty()).
It looks like I can push these changes straight to your master since it is
a fork of mine.  I notice one of my changes was already in your updated
pull request.  The comments should be on the GitHub file (ProcessSong) if
you needed to see them.

It all looks really good and thank you again!  I've merged the pull request
into the master branch having checked the changes in the code and tested
the output on my devices.

Cheers again for your hard work!

Gareth

iv-gha:
Hello Gareth,

Well that sounds better - thanks.  I was confused as the build had worked.
There is always a risk when making change so well done for finding a fault
during your testing - please do more!  You may like to beta for a while to
hopefully catch any more!

On a side note -I tried the revised connect  On the old version it sent the
song over the link becoming a received file? - this meant the receiver did
not have to have the song at all.  For me this is a strong advantage
allowing easy share of a song and changes as the song is rehearsed.  The
change to nearby seems to depend on having matching song libraries?  I
think sending the song is the much stronger design - very flexible.
Hopefully useful feedback.

I am happy to add my 'bits'.  Thanks for the tips.

Kind regards
Ian V

thebigg73:
Hi Ian,

The old version (Salut) was using an abandoned library that was throwing
errors that I couldn't fix and was becoming temperamental with
connections.  It also relied on WiFi direct, which not all devices have.
The Nearby API is far more flexible and reliable, using a mixture of
Bluetooth, Bluetooth LE and WiFi as required.  It also seamlessly switches
these services on and off in the background as required.

Both old and new send the song file. The old method had to include a 'hold'
pattern before sending (and a hold delay before processing) as it often
sent the information several times in quick succession - that doesn't
happen now.  I think this was partly down to timing issues of multiple
devices loading the song.  To send the file, one device has to act as the
host (toggle button in the menu).  If a device acts as a host it sends the
OpenSong files as encoded text at the same time as a standard link
(folder/filename pair), swipe direction (L2R or R2L).   If the other user
(not the host) is connected, they will receive all of the information and
their settings decide what to do with it.  If they haven't switched on
'receive host files', their device uses the link and relies on the matching
file (and folder) location to open their version of the song.  If they have
switched on 'receive host files', they will decode the song into a
temporary location (it goes into the 'Received/ folder and a file called
ReceivedSong is created).  This is what is displayed on their screen - an
exact copy of the host file.  This doesn't use their own version of the
file - it uses the host version instead.  If they have also switched on the
'keep host file', this will then create the song file in the matching
directory location (it should create the folder if it doesn't exist as
well).  It will overwrite their own version of the file, or create a new
file if they didn't have it.

If the host uses a non OpenSong file (PDF/Image), the song link is still
sent as a byte payload instantly, but instead of encoding the song with
this it as bytes and sending directly, it sends it as a separate file
payload.  The bytes payload has a maximum of 32kb - more than enough for
OpenSong text files (although if they are bigger due to embedded images
from the desktop app, the app checks this and sends as a file instead).  If
the user (non-host) receives/keeps host files, the app waits until the
transfer of the pdf/image is complete before processing it (the host
carries on as normal).  It uses the link sent firstly as bytes as a way of
getting ready to create the file if the user has selected 'keep host file'.

The nearby feature (as did the old method) includes the option to send an
autoscroll start/stop command from the host to connected devices.  It might
be worth a bit more checking of that.

Regarding bug testing, I suspect that given the Covid19 situation, the app
is getting far less use than normal just now.  When I used to release
updates I tried to do these on a Mon/Tues to allow testing before the
weekend gigs. Most bugs/issues were only flagged up on Friday-Sunday and
little during the rest of the week.  Beta testing has a disclaimer and it
is easy to unsubscribe from this, so it is quite good for testing of
potentially big issue bugs (i.e. ones that cause crashes that can be
replicated), knowing you can revert to the stable channel before a gig.
However, there are only about 360 testers registered currently, compared
with 13k+ active users (those using it more than once a week), so smaller
glitches like this tend to go unnoticed in beta testing, but do get picked
up by the stable release.  Most issues are easy to fix quickly and I don't
mind doing updates this way, especially since it is free (I could've
retired if I charged £1-2 per install!)

Given that we know the changes made, we can target our testing around these
parts of the code (and have done as we make the changes in our own
branches).  I'm reasonably confident that no show stopping bugs have made
it through so far!  However, you've found some bugs in my code that have
gone unnoticed for over a year!!  Most people just don't notice or assume
it was a glitch caused by something they've done and don't report it,
especially if it doesn't break the app.

I'll do a bit more testing today and push the changes up to beta.  You'll
notice that the manifest file will keep getting updates to the version code
and version name - Google Play requires a newer version everytime you
submit an update.  The next one will be 180/5.2.0.  Just in case you wonder
why that keeps changing.


Best wishes,
Gareth

iv-gha:
Thanks Gareth,

Great detail.  Things to think about.

I am not seeing the received file mode kick in when I have 'received host'
song on.  I add a new song on the host and display and the result in the
'not on device' response.  It does not seem to be behaving in the 'receive
host song' mode.  App restart did not help. v5.1.8. I am using the current
dev build on the other though.  Please consider checking and also
perhaps that it works well with variations?

Kind regards
Ian V

thebigg73:
I'll have a look.  Because the old version and new method aren't
compatible, both devices need to be running V5.1.8+

PR #75, opened 2020-08-05
For consideration: A song info block 'extra information' block and right trim of song lines

iv-gha:
Hello Gareth,

I have been looking at exports and wanted to add a 'song details' block (as seen in txt outputs) to png and pdf exports.  This ended up being a song details block for on-screen and export use.  I think it also happily gives an option for users who want a more traditional on-screen printed page look (maybe).

I have also looked at right trim of lines for screen display (helps to maximise size and compensates for untidy imports).  I am less sure of the utility of this.  It did however allow me to put 2 spaces after lines - to reduce the right edge overruns that I occasionally see.

Hopefully something useful.

Kind regards
Ian V


iv-gha:
I will redeliver with fixes.

PR #76, opened 2020-08-06
Song details extra information for exports

iv-gha:
Hello Gareth,

I have been looking at exports and needed a 'song details' block (as seen in txt outputs) for png and pdf exports. It is in the 'extra information' mechanism.  It then appears on screen for export use.  It gives a more typical printed page look (maybe).  I add 'notes' at the bottom of the block - but they could stay top right as now.   This version splits long lines to try to ensure good multi-column song display.

The 'song block' also gives users the option for a more printed page look on-screen.

I also right trim and add 2 spaces of lines for on-screen display - this give maximum size for 'untidy' files and reduces right edge texts overruns that I occasionally see (Text overruns would be better fixed elsewhere?).

Hopefully something useful.

Kind regards
Ian V


iv-gha:
Having looked at  styles for printed songs, I think using bold is better than all capitals for the title.  It is still obvious and is not so wide.

I have just coded an override for lyric lines (taking a similar approach to ;__ for comment lines), I propose a prefix with 'underline B underline' to get bold on a lyric line.  I will see how it behaves!

Added: Here is a png export, using a black and white theme, printed to PDF using Xodo.  Quality is good?
[All heaven declares.pdf](https://github.com/thebigg73/OpenSongTablet/files/5040866/All.heaven.declares.pdf)

Kind regards
Ian V

iv-gha:
Hello Gareth,
There needs to be a liitle work to make it work better with stage mode!

thebigg73:
That image looks good Ian!  I'll have a look.  I switched to using Android built in PDF writer recently (I was using a third party library that I've now removed).  The image/screen capture can be embedded as a PDF in the app.  That's how they used to work.  In the recent versions I've used the pdf writer to create a proper text based pdf, but only using monospaced font for now to avoid the spacing issues.  There's no reason why we couldn't plan to write directly as a PDF in future.  The screenshot version was my quick easy of getting an version of the nice screen layout.

The benefit of a proper text based pdf is much reduced file size and scalable content (no loss in quality when zooming in).  The screenshot is a compromise of Android memory and image file size compromise.

Your display does look good though.

I've not had a chance to pull it in to test yet

thebigg73:
The text overrun is due to scaling the font size - it depends on the font being used - some fonts are better than others.  I've sorted this on the material app branch by scaling the layout instead and drawing it once (also massively improving the drawing speed).  It definitely gives 100% maximum sizing now, so I'm not too worried about it in this version (remember I'm pulling all the good stuff you've added into this as well as I go - just about to start bringing in the auto scroll and pads next!).

iv-gha:
Hello Gareth,

The image scaling of the inbuilt pdf seems to let it down - very blocky.
The approach I took was to use png and print externally.  At the mo this
gives useable quality / size.  Even under zoom it is not too bad.  Whilst a
full text pdf would be good I note that a lot of PDF out there use image.
For me avoiding a further render engine is a huge benefit of your approach.
The user also gets what they see. Now if opensongapp could compress and pdf
the png!

Getting a good layout is my focus right now.

Have fun.

Kind regards
Ian V

iv-gha:
Hello there,  as regards the overrun...  Yes, I saw a note in prentation code about material scaling - sounds good.  I essentially do what you did there - I found your 2 spaces note after I added elsewhere.  I guess we have some goodness to spread!

I have excluded notes etc from the text based exports.  Stage mode also has these to be dealt with in some way.  I also noticed that when songs are in presentation mode 'extra information' are not available - intended?

Kind regards
Ian

iv-gha:
Hello Gareth,

Fix for stage mode splitting up the song details block is to  change 1750  line as below.

1748        if (StaticVariables.whichMode.equals("Presentation") || StaticVariables.whichMode.equals("Stage")) {
1749            // IV - Song block dividers ||| are not processed as splits
1750            song = song.replace("|||","|").replace("||", "%%LATERSPLITHERE%%");
1751        } else {
1752            song = song.replace("||", "");

I am not sure how stage mode would display the song details section - my devices do not do screen share" - hopefully good.

Kind regards
Ian V

thebigg73:
Hi Ian,

I've been enjoying the last few days of sunshine before going back to work
tomorrow (I'm a secondary school science teacher by day), so haven't had a
chance to test out your code yet - sorry!
I'm obviously losing my mind somewhere.  I though I'd added code to create
text based PDFs of songs in the last few updates (rather than using the
embedded image method).  I had even created A couple of songs when testing,
but I now can't find out how I did it (the export method in the app still
uses the embedded image method).  I'm wondering if I had only switched it
on in my MaterialApp version.  I'll test and get back to you!  As a result,
you are probably unaware of what I was talking about in the last email!
From memory, the pdf/PNG export of songs was only an option in the
performance mode as it required the full song to be rendered to the screen
in order to capture the image.

iv-gha:
Hello Gareth,

No problem at all.  The song details block was added for export, however I
quite like it on all the time as it gives a good printed page look.  I
realise there is duplication of title and author with the top bar. However
with top bar hidden and buttons reduced to one it gives a very printed look
which I like (even though the block  takes up some of the screen). Take a
look when you are able.

I hope work goes well.

Kind regards
Ian V

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
```diff
@@ -2622,6 +2621,14 @@ private void customButtonLongPressAction(String s) {
                 }
                 break;
 
+            case "autoscale":
+                // IV - Take the current value, lookup in string and set to the next value in the string
+                preferences.setMyPreferenceString(StageMode.this,"songAutoScale", "" + "YWNY".charAt("YWNY".indexOf(preferences.getMyPreferenceString(StageMode.this,"songAutoScale","W")) + 1));
```
thebigg73, line 2626: Nice idea!

app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
```diff
@@ -5245,6 +5251,10 @@ protected void onPreExecute() {
             try {
                 // We know how many columns we are using, so lets go for it.
                 column1_1 = processSong.preparePerformanceColumnView(StageMode.this);
+                // IV - If doing song block - needs own preference
+                if (!(preferences.getMyPreferenceString(StageMode.this, "stickyAutoDisplay", "F").equals("N"))) {
```
thebigg73, line 5255: No issues adding another preference.  Add it to the comment lines in the Preferences.java file (allows tracking of the variables currently in the app).

thebigg73:
Hi Ian,

One potential issue I've come across:
I know of a few users who run stage mode on their 'on-stage' devices that connect to a second device running Presentation mode (in order to have nicer backgrounds, etc.).  The idea is when they click on a section in Stage Mode, this mirrors the sections produced in Presentation Mode meaning the projector shows the same content they are using on stage.  Currently with the block display, this 'extra' block/section appears in Stage Mode (and Performance Mode), but not in Presentation Mode meaning section[0] in both modes points to different content (may also throw a caught out of bounds error when Stage mode device clicks on the last section which doesn't exist in Presentation Mode - a quiet try/catch deals with, but the sections aren't in sync.  Given you are more up to speed with how this is generated, is this an easy fix for you?
I'm now realising looking at the code, that this would already be an issue before (with my code) if the user had switched on the 'Next song in set' extra info to display at the top of the song.  It defaults to be at the bottom of the song, so would have been dealt with with the quiet try/catch block in Presentation mode, but the song would be in sync on both devices up to that point.
Any thoughts?  We could just accept this issue, or warn users about it in the user guide, however, it would be nice to fully sync this up between app modes.

Other than that, it looks great!

I can only direct edit files you've included in the pull request, so couldn't just edit your version.  I'll create a pull request back to you from my test branch with the new variable.

iv-gha:
It would be easy to restrict extra info to performance. A tweak to the ifs
in the extra section. Complete removal will keep stage and presentation
mode in step (where devices have matching song files) and avoid the silent
fails.

Will users miss extra info in stage mode?

Kind regards
Ian V

thebigg73:
That sounds like a sensible option.  Stage mode isn't supposed to represent
printed music.  Often it's landscape orientation and a chord version of
what would be shown in church for people to sing along with (verse or
citrus at a time) with a sneak preview of the next section below it.  Given
the export function as well (for pdf/PNG) isn't available in
stage/presentation mode, let's keep these in sync and only allow the block
info for performance mode.

thebigg73:
Hi Ian,

I've come across a glitch with the OpenSong/Chopro conversion in the edit
song window.  I think it was likely an issue before, but given your track
record in making this work far better, any thoughts on this?
I've attached a copy of the song.  Edit it and try alternating between the
chord formats to see the issue (click on the X to close without saving for
future tasting).

Best wishes,
Gareth

thebigg73:
Hi Ian,

I'm going to pull this into the Master branch.  We are aware (with a potential fix) of the issue between Stage/Presentation Mode and the OpenSong/ChordPro conversion (which is minor and most people won't see the issue), again I'm sure we can sort this.

Well done again!

iv-gha:
Hello Gareth,

Can you paste the file into a reply please.  I cannot find the attachment.

Chords on their own have always been a problem for me,  finding a format
that is stable on convert is an art! The use of | and ||:  and :|| is a
problem requiring care.  I will be interested in how people format.

Kind regards
Ian V

PR #82, opened 2020-09-24
 Fixes and feature improvements : Autoscroll and pad start, nearby song sends, CCLI logging, presentation mode bahaviour

iv-gha:
Hi Gareth,

I have tweaked quite a few things and am thinking you might like a look!

The biggest by far is for Presentation mode - I got a Chromecast!  I think I have got stable fades, stability when under stress and better action handling.  This has (hopefully) improved display behaviour.

I introduce features to (maybe) optimise nearby song sends.  This was to stop obvious stress on aged client devices during rapid song changes and to give a high chance for a sync start of autoscroll with clients.

Apologies for lots of layers at once (I think you wanted a break).  Hopefully of interest. You may have questions.

Kind regards
Ian

added:  I have noticed infrequent animation stall / screen update stops on Chromecast - very irritiating!  Perhaps this is the reason for the 'panic'?

### DIFF
--- a/app/build/generated/source/buildConfig/debug/com/garethevans/church/opensongtablet/BuildConfig.java
+++ b/app/build/generated/source/buildConfig/debug/com/garethevans/church/opensongtablet/BuildConfig.java
@@ -7,6 +7,6 @@ public final class BuildConfig {
   public static final boolean DEBUG = Boolean.parseBoolean("true");
   public static final String APPLICATION_ID = "com.garethevans.church.opensongtablet";
   public static final String BUILD_TYPE = "debug";
-  public static final int VERSION_CODE = 182;
-  public static final String VERSION_NAME = "5.2.2";
+  public static final int VERSION_CODE = 183;
+  public static final String VERSION_NAME = "5.2.3";
 }
--- a/app/release/output-metadata.json
+++ b/app/release/output-metadata.json
@@ -11,8 +11,8 @@
       "type": "SINGLE",
       "filters": [],
       "properties": [],
-      "versionCode": 182,
-      "versionName": "5.2.2",
+      "versionCode": 183,
+      "versionName": "5.2.3",
       "enabled": true,
       "outputFile": "app-release.apk"
     }
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -2,8 +2,8 @@
     xmlns:tools="http://schemas.android.com/tools"
     package="com.garethevans.church.opensongtablet"
     android:installLocation="auto"
-    android:versionCode="182"
-    android:versionName="5.2.2">
+    android:versionCode="183"
+    android:versionName="5.2.3">
 
     <uses-feature
         android:name="android.hardware.touchscreen"
--- a/app/src/main/java/com/garethevans/church/opensongtablet/CustomAnimations.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/CustomAnimations.java
@@ -145,6 +145,7 @@ class CustomAnimations {
             @Override
             public void onAnimationEnd(Animator animation) {
                 v.setVisibility(endVisibility);
+                v.setAlpha(endAlpha);
             }
 
             @Override
@@ -189,6 +190,7 @@ class CustomAnimations {
             @Override
             public void onAnimationEnd(Animator animation) {
                 v.setVisibility(endVisibility);
+                v.setAlpha(endAlpha);
             }
 
             @Override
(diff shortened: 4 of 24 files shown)


thebigg73:
I haven't forgotten!!!

I'm tied up with work stuff just now, but I'm hoping to get a look at this over the weekend.
Thanks again Ian!

iv-gha:
No problem.  There have been some further tweaks, so in this case later is
better!

Kind regards
Ian

thebigg73 (approved):
Initial skim through code looks fine.
Will merge pull request and test.

iv-gha:
Thought - Thinking on CCLI on screen info - to be legal this should be the full text.  As this varies I think code some entry fields on the CCLI setup page! Users (you do not what to pick through CCLI terms for all countries!) can then add what makes them legal.  Is this sensible?

For the stalls in animation on Chromecast...  I noticed a motion background reduces this, so tried a manual animation at end of auto animations to see if it helps - it does! Stalls still happen but the manual animation helps and it allows panic to also run.  My tests simulated a poor connection which had many stalls before this tweak.  Not a total cure but much better.  A bit messy though.  Do you have more info on Chromecast stalls?

thebigg73:
Regarding CCLI - not sure what the other stuff would be.  The output of this doesn't go to CCLI, but is a log to help complete their return.  I'm basing it on the last return I was in charge of (but that was a few years ago!).  If you've further info about what is required, let me know - it should be easy enough to add.

The Chromecast stalls I've found were simply down to poor WiFi connection.  It doesn't need internet access as the app acts as the server, but does need a good connection.  Connecting via HDMI fixes all of the lags.  I added the panic to try to make sure everything that was facing in was fully shown after a few seconds in case the animate in died.

iv-gha:
For CCLI it is the copyright message that needs to be displayed on projection I am thinking of - at the bottom of the screen.  The format is prescribed and includes the CCLI licence detail.  I have gone someway towards it in my code.  To be legal users need to substantially follow the format and show at least once during a song.  Opensong app shows something all the time but to be assured of legality it should show more and the easiest option is to fully meet CCLI rules.  I indicated in the code where it would go, I think user fields are needed as it would be onerous to try to source the correct words for every country.  It would at least enable users to be legal if fields where available to adjust the text.  I would choose to be legal.

I will put up an 'ish' fix for poor Chromecast on my branch.  It will not hurt to help those like me who are without an HDMI output. Even with good wi-fi I see the occasional stall.  Modern tablets of any major brand seem to have abandoned HDMI! All I can find are slow ones!

iv-gha:
CCLI say notice must substantially be like

“Hallelujah” words and music by John Doe
©2000 Good Music Co.
CCLI License # 0000

So for opensongapp a close approximation would be

"[song title]"
[New field song author prefix] [song author]
©[song copyright]/n
[New field CCLI prefix] [CCLI number]

If the system has a CCLI number and the song has a CCLI number then format details in CCLI format with license number displayed.

Might even want a field for © as a prefix too.

For UK [New field song author prrefix] would be 'words and music by' and [New field CCLI prefix] would be 'CCLI License #'

iv-gha:
I have put up an 'ish' fix  - it does small adjustments which appear to (nearly always) kick chromecast back into life.  Does not appear to cause problems - shame on google for it being needed!  I can pull - see if this is good.

     @Override
            public void onAnimationEnd(Animator animation) {
                v.setVisibility(endVisibility);
                // IV - Give the end Alpha a manual animate - this is done because cast animations can stall
                // IV - We do not assume the animation completed - we adjust the current Alpha value
                float kickAlpha;
                kickAlpha = v.getAlpha();
                if (endAlpha == 0.99f) {
                    v.setAlpha(kickAlpha + 0.01f);
                    v.setAlpha(kickAlpha);
                    v.setAlpha(kickAlpha + 0.01f);
                } else {
                    v.setAlpha(kickAlpha - 0.01f);
                    v.setAlpha(kickAlpha);
                    v.setAlpha(kickAlpha - 0.01f);
                }
            }

thebigg73:
Ah!  I thought you were talking about the CCLI reporting log, not the projected details!
I've tested your changes so far (without the last stuff you've mentioned) and it looks really good!   I've put a couple of changes in for initiating Casting (updates via Google) - do not affect your code.
Also, I've had a few users commenting that autoscroll times are now way out (songs taking about 10% longer than specified).  I've found the issue - it was the checkCanScrollDown/Up code ignoring the newPosFloat value set by the autoscroll method.  It instead grabbed the value from the current scroll position (which is an int and not a float).  This effectively rounded down each scroll movement and caused the scroll amount to be smaller than it should have been.  I simply added a test for isautoscrolling before grabbing the value (autoscroll set this during its logic).
I've pushed all of my changes back to the master.

iv-gha:
Sounds good.  Sorry for the CCLI confusion.  I have merged your updates to my branch - thanks.  Ideally the presenter mode changes just make it feel well behaved!

PR #84, opened 2020-10-09
Presentation mode - Improved PDF support and simplified infobar display

iv-gha:
Hello Gareth,

I fixed a crash related to PDFs and this revealed PDF support in presentation model (hidden by the bug)!  This led to fixing a few things for better PDF  support.  Please consider.

I have also looked at only having infobar once during a song, this meets CCLI rules and reduces clutter.  I have done this with this pull.  The approach happily simplifies infobar fade activity from 5 to one!  I would appreciate your views on this.  Hopefully good stuff.

Commit comments have more detail.

Added TODO - Need to rethink Alert as it will now also be hidden! Whoops.

Kind regards
Ian

p.s. The icon to the left of infobar text for stage and performance mode no longer displays - looks like a recent compat change to cast_screen.xml has not quite worked?


thebigg73 (approved):
PresentationCommon changes - notice the use of Boolean rather than boolean.  Not sure if this is required (Object instead of primitive).  Slightly less overhead using primitive boolean, also I don't think null is required, so will look to change this and check it is ok

thebigg73:
I'm finding the infobar only hides sometimes

thebigg73:
I think this was due to the 10s delay.  However, if moving to the next slide in the same song, the info bar shouldn't be redrawn (it currently animates out and in with the song lyrics). Normal behaviour is that no redraw should be carried out if the content hasn't changed.  I'll try to fix.

thebigg73:
Ok, added this feature and pushed back to the master branch.  Let me know if this is better.

iv-gha:
The 10s mandatory display is deliberate.  A user may quickly flick to another slide without good (definitely legal)  info display. Also it appears on the first (any) section shown after a song change or return from logo/blank.  With update projector off, it is possible to start in the middle of a song. Sometimes starting in the middle is what is intended.  I may of course be over thinking it!

I agree that the block should remain on screen if the user moves quickly to a different section. But only stop when shown continuously for 10s or more.  My feeling is that once selected the proper use of a section will be longer than 10s - so no display for next used section.

There remains the issue of alerts which I am looking at.  I have also noticed lyric text positioning for 'top' has text too close to the top! I think this is a bug and not caused by the infobar display change.

It not quite ready to go.

Are you good with the principle of display infobar 'once' and is the PDF stuff looking good?

Thanks for the Boolean and boolean thing.  I must try to actually learn the language! :-)

Happy to be involved.

Kind regards
Ian

thebigg73:
Definitely good with your changes.
The updates I've made will keep the infobar displayed for 10s, but I've stopped it fading in/out when moving between slides.  After the 10s, it will then disappear.  It's just the fade in/out of the infobar I didn't like while it was still showing within the 10s limit.
Regarding the alert - it might be worth just moving it out of the infobar section as a stand alone view.  Info bar could position below it?  If it isn't used, it will be set to View.GONE anyway.  The content would then be set to display above the alert text rather than the infobar.

Regarding the spacing, I'm pretty sure this is down to the TV sizing.  You can adjust the vertical and horizontal margins on the display layout to fix this.

iv-gha:
Yes not fading is good.  I can see the changes but am in the middle of looking at the alert - I am thinking to keep it at the bottom (the traditional spot!) rather than split the song.  This is perhaps an easier 'sell' for users.

Agreed on the separate view as it has different timings to infobar now. :-)

Kind regards
Ian

PR #85, opened 2020-10-11
Presentation mode enhancements

iv-gha:
Hello Gareth,

I gave a separate Alert bar a go but quickly came to the conclusion it introduces too many code overheads (it is the nature of the common code approach) and doubles up most processing for infobar.  There is already lots of code to maintain!

I have worked up a change to have Alerts continue to share the 'infobar' with the song info - take a look.

I have combined fade out and fade in into the same process - so this reversed the need for some of the string routines that were added as the code only runs once (I felt a bit odd reversing some of your good work - apologies).

Alerts could appear with song info - meaning a fade from song info with alert to just alert on the next section change.  My code deliberately stops this by making it either or. So song info alone, followed by alert alone at other times.

The Alpha animation 'kick' need to be less aggressive.

Some other smaller fixes / adjustments are included.

Your further thoughts would be appreciated.  Have fun!

Getting there.

Kind regards
Ian

p.s. Please ignore the extra files! I accidentally pushed more that just the Java and Cast.xml files (darn you github!)
p.p.s Sorry for the multiple commits!


iv-gha (commented):
Please ignore mylibrary and build files - these are not intended as part of pull!

thebigg73 (approved):

thebigg73:
Hi Ian,

I like your changes and I think I'm happy to push this update (especially to the Amazon App Store) to get functionality back.  I can't believe I never thought about the fact that I had the Play Store installed on my Fire 7 that was masking the issues (because everything worked fine).  Well done on spotting that!
I've included links on the app for affected users on how to install the required files.  Tested it all with my Fire 7 (7th gen) and it all works.  If you get a chance to test it with the Fire HD10, even better!
I'll push my changes up to the master branch now and get ready for packing for the app stores (I need to do some translations first).

Cheers,
Gareth

iv-gha:
Sound like we are there.  I still have a friend's fire 10 so will quickly test...

iv-gha:
Hello,
Yes the changes are going on fire hd 10 7th Gen.
Comments
- The snack message for nearby is only displayed a short time - doh - too short?
- The cast icon appears but does not work until play is installed.  Anyway to suppress or give a message on use when broken.  Sadly it is not broken enough when broken... I could see and select my Chromecast only to get no response. I might think it to be opensongapp at fault.  However, it does get fire users working - great.
Thanks for the hard work.
Kind regards
Ian

thebigg73:
The snack bar only allows short, long and indefinite.  I set it to long as
I didn't want it to be a constant nuisance.  The Snackbar triggers on first
run for the cast issue (there is a more permanent message on the boot
screen which shows after the app updates).

If the users click on the connect menu item it shows the Snackbar again.

They also see the message when opening the fonts settings and the choices
are hidden.

Gareth

PR #91, opened 2020-11-17
Menu xml and search adjustments

iv-gha:
1. Main menus - use of right arrow button icon with background colour the same as menu
2. Song menu - use of left arrow button with background colour same as menu, button re-positioned to left

Search songs
1. Now displays 'folder/filename' instead of just folder, with this at the bottom. java and xml changes
2. No longer displays folders in song list

For CCLI use the song title needs to be the authorised title.  Songs are ordered by filename in the 'Song menu', having filename visible during search is helpful.  For example, I use filenames of form '[Part of first line] ([Song title])' as it is common to refer to songs by first line - having the filename shown helps to select the song with the right [Part of the first line].


iv-gha:
Hello Gareth,

I've had a chance to look at using arrows on just the menus - turned out that everything is in place to do this simply with xml changes - great.  My thinking is. if we are changing menus then to include this with the other menu stuff.

I also tweaked/fixed the song search - minor stuff but I think useful.

Kind regards
Ian

Added:

1. A 'strings' update as capitalisation changes are needed.
2. A proposal for layout changes for the 'Toggle autoscale' page.

Comment: There is a slight button shadow effect for the arrow buttons! Probably should be implemented as just the icon.

thebigg73 (approved):
Looks fine to merge and test
Cheers,
Gareth

thebigg73:
Something very strange happening with the song/chord display.  Most chords seem to end up being pushed to after the entire lyric line?

thebigg73:
Found the issue - in ProcessSong if displayChords is false, the app was joining up the lyrics text.  However, unless the user had set this value the check was defaulting to false, whereas it should have defaulted to true (as this matches the default display mode).

iv-gha:
Good catch - had not seen that!

PR #101, opened 2021-01-09
SongSelect Lyrics only import - rework to support all languages + fixes/enhancements

iv-gha:
Hello Gareth,

usr extract changed to extracting from HTML.  This supports all languages and supports any headings present as tag (and we are not fussed what they are!).  I have applied the chordpro approach to usr!

If you think HTML scraping is reliable for CCLI usr style pages then this resolves support for all languages.  A win!

Have a look. 

ADDED: This includes handling multi-line copyright notices, as this was very evident in 'Lyrics only' format.  I can now see that this was a problem for chord sheets (only using the first line) before - so hopefully fixed for both!  Might this also happen for authors (trying to find a very long author line!)?
(FIXED - See commit.  Versions of Silent Night are useful to see the fix working)

Kind regards
Ian


iv-gha:
This pull has expanded to include further work on the Metronome and a fix for a crash.

Kind regards
Ian

thebigg73 (approved):
Pulling these in for testing.

PR #103, opened 2021-02-11
BootUpCheck.java proposal / Song delete change / Metronome timing / Transpose improvements

iv-gha:
Hello Gareth,

BootUpCheck

Here is a 'quicktips' version of BootUpCheck.  This is trying to keep the initial screen clean.  The extra advice is now Quicktip info displayed using an (i) button.  I think you might have wanted to move in this direction.  It helps smaller screen devices. The user manual (?) page will need tweaking to reflect - if we go this way.  Hopefully the buttons are positioned sensibly!

I have added a block of paths that start '\tree\msd:' to try to stop use of 'Downloads' on Android 10+, it was already blocking '\tree\raw:', can you also consider if this is safe to do.

Metronome timing

I have added a metronome update - I managed to re-shape the loop to clock based beats!  The last update reduced (but did not eliminate) the risk of a drift from the visual metronome.  I think this clock approach could be more robust.

Transpose improvements

I have tried to correct for the creep of chord positions over lyrics due to the transposed and original chord having different numbers of characters.  It is not perfect but I hope much improved.  Comparing to the old behaviour is probably a good way to see the benefit.

Thanks for considering these bits.

Kind regards
Ian


thebigg73 (commented):
Some really good fix suggestions in here!  I can't believe I never thought of the system time match for the metronome!
I'll need to look into the tree/msd: bit - can you give me more info on your reasoning/thoughts - I'm struggling to find info online about what it is or when it occurs.
Regarding the connection restart - I'll test, but ideally once a connection is made, connected devices should be able to switch between host/client without needing a reconnect.  I think your code doesn't change this (just stops the advertising/broadcasting). What I/we will need to fix is some check that if the user when initiation connections enables nearby connections and then switches on act as host, it won't work currently as the nearby connection starts in client mode.  A sanity check on changing host mode should be simple though (if not connected and host mode changed, stop searching and start broadcasting).  I'll sort this.

Many thanks again for all of your input and well done for deciphering my old code!

iv-gha:
Hello Gareth,

On my Android 10 device if you select contents of the downloads folder it returns something starting tree/msd: - my feeling was that even if things would work if selected from the downloads folder (and sometimes they will!) we want to be consistent and say no to the use of downloads.  tree/raw: is seen on earlier Android versions.

The logic for the kick on change of host mode is to do the logic of a stop and start of 'Enable nearby'.  I think I based it on that code.

Transpose is, I think, much better behaved.

Have fun!

You also got a few other bits.  Removing comments from lyrics on presentation is a chordpro behaviour I think.

Kind regards
Ian V

iv-gha:
Hello Gareth,

Can you consider allowing nearby connect to reestablish a nearby connection without prompts? So if a session has been agreed previously (using prompts) and it disconnects and tries to reconnect with same host then accept without further prompt. 

I am using stage mode connected to another device in presentation mode to show words.  If the session collapses the devices reconnect and prompts appear - I am playing guitar so would prefer that they sort themselves out without prompts. I think nearby can do prompt-less connects and maybe should in this case?

For those one man band moments!

Kind regards
Ian

thebigg73:
Hi Ian,

I don't think that's possible.  It's pretty of Android drive to more
security that you have to give permission.  The 'check the code matches
statement' is optional (but recommended), but you still need to confirm
connection again, even if you've previously given permission

thebigg73:
Should have said, you can connect on your stage mode device and present
that way - I'm sure you knew that though.

iv-gha:
Hello Gareth,

I have had a quick look and Android Developer info for Nearby describes
methods for re-connection, with the fairly obvious need to keep a record of
connected sessions so that the session can be accepted without prompt.
With Wi-Fi solutions disconnection is part of the problem, it is a shame
that nearby itself does not handle it - but it appears Devs can code to
handle.   The ideal is no prompt at either end.   A dropped connection will
currently interrupt the host (and client) on attempted re-connection -
having a dialog appear in front of your song-sheet when playing
guitar/singing is less than helpful!

I have a tablet with HDMI output and so can run this in stage mode.  It
would need the option to use presentation format 'slides' for HDMI.  So,
see the lyrics/chords, control slides using stage mode sections, present
slides over HDMI in presentation format.  Can we give Stage mode this
option?   My problem then becomes running HDMI cable from the tablet to the
AV distribution point (no use of unstable Wi-Fi!),

This is the one-man 'band and AV team' scenario.

Thanks for considering.

Kind regards
Ian V

iv-gha:
Hello Gareth,

Looking at the code I think a way forward is to have a Stage Mode option to display only lyrics (ignored when chords are displayed).  I have crude code that suppresses the relevant lines to give a basic presentation style. 

Kind regards
Ian V

thebigg73:
Isn't that already an option?  I know that performance mode can display
lyrics only or lyrics and chords despite what the tablet mode is showing.
It's an option in the connected display.

I've had a look at the reconnect code looking for the device name and
reconnecting if it was previously found and it seems to work on the device
that I'm running the test version on.  The other connected device is
running the production version and still shows the prompt if host or
client, but the newer version just works.  I currently can't install on
another device due to my usb port misbehaving and my other device not
allowing ADB over wi-fi (so frustrating!).  If you've got two devices you
could check, that's be great.  Connect the devices, then on one switch off
nearby and then switch it back on again

iv-gha:
Hello Gareth,

Nearby stuff sounds good I will have a look.

For the stage mode, we do not have stage mode section by section control of a presentation style HDMI output on one device. To get this you have to connect a device in stage mode to a device in presentation mode to control using stage mode and connect the output of the presentation HDMI/cast .  One device alone cannot do this (at the moment).

Kind regards
Ian V

PR #104, opened 2021-02-27
Metronome - Forgot to handle 6/8!

iv-gha:
Hello Gareth,

When using the app tonight...

Apologies. I broke metronome 6/8 time.  I have fixed.  I calc durations starting from a 1/6th of a beat.  Why? This avoids  rounding issues. For 6/8 this guarantees that 3 lots of the sound at 2/6 beat exactly matches 2 lots of the visual at 3/6 beat.

Kind regards
Ian V


thebigg73:
Glad is isn't just me that does stuff like that!

PR #105, opened 2021-03-05
Nearby tweaks

iv-gha:
Hello Gareth,

Trying to help nearby behaviour with a focus on preventing a user being disturbed when connections are made...

1. Host must be intentionally in 'Connect Devices' to respond to new connections.
Aims - No prompts appearing on screen when the host user is performing.  Intentional management of new connections.

2. Client is not interrupted when another client connects to the host.  The client will receive the 'general' open song request but ignore when the current song is the same.
Aim: To support another clients re-connection during performance, that client will be known by and silently connected by the host.  The use of the app on existing clients should not be disturbed.

Please consider. Getting there!

Kind regards
Ian V

p.s. Some other minor tweaks

### DIFF
--- a/app/release/output-metadata.json
+++ b/app/release/output-metadata.json
@@ -10,8 +10,8 @@
     {
       "type": "SINGLE",
       "filters": [],
-      "versionCode": 190,
-      "versionName": "5.3.0",
+      "versionCode": 191,
+      "versionName": "5.3.1",
       "outputFile": "app-release.apk"
     }
   ]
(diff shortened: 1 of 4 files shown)


thebigg73 (approved):

thebigg73:
I've made a few tweaks by adding a preference for the 'listen for clients when menu is opened/closed'
Also added a bit of changes to the logic.  Switching on host mode if not already connected, switches on enable nearby.  If nearby was already on, it changes discovery to advertising.
Note that once a connection is made (and enable nearby stays on) devices can switch between host/client mode as required.

Let me know if this seems ok and if we should push to the Play Store?

I'm wondering if in the bootupcheck screen we should just remove the (?) button and instead just add a link in the (i) popup text to say something like 'click here to find out more' which then hyperlinks to the website example.

iv-gha:
Hello Gareth,

I will have a look this morning.  I too had found the change of mode a problem and had changed to just clicking and un-clicking the nearby button in the code!

I agree on the (i) popup point,  this will make the change subtle but hopefully obvious enough for those having trouble.

Kind regards
Ian V

thebigg73:
I've committed my updates suggested.  I like what you've done so far with
the idea of silent connections and stopping unnecessry song reloads

PR #106, opened 2021-03-06
Nearby testing tweaks

iv-gha:
Hello Gareth,

Thanks for your patience - it takes a while to test Nearby!  I discovered metal cake tins are good Faraday cages, a lot easier than walking up the garden to test disconnect / reconnect!

I think these few tweaks get us to the point of being happy to release 'silent' reconnect Nearby.  Please consider.

Hopefully all the other stuff is good?  Transpose is, I think, much better behaved.

Yes, please do change BootUp to just have (i).  I am happy to strongly point to Android being the pain and not the App!

Kind regards
Ian


iv-gha:
Hello again!

A quick thought... if BootUp select returns an invalid path then perhaps we open the (i) info for the user?

Kind regards
Ian

app/src/main/java/com/garethevans/church/opensongtablet/NearbyConnections.java
```diff
@@ -249,12 +253,16 @@ public void onDisconnected(@NonNull String endpointId) {
...
-                // Try to connect again after 2 seconds
-                if (!StaticVariables.isHost) {
-                    Handler h = new Handler();
-                    h.postDelayed(() -> startDiscovery(), 2000);
+                if (StaticVariables.isHost) {
+                    // Check if we have valid connections
+                    StaticVariables.isConnected = stillValidConnections();
+                } else {
+                    // Try to connect again after 2 seconds
+                    if (!StaticVariables.isHost) {
```
iv-gha: Sometime you do not spot the obvious - please remove the unneeded 'if (!StaticVariables.isHost)'.  Compiler will sort this kind of thing but hey!

thebigg73:
Hi Ian,

I've made the changes to the (I) on bootupcheck.  To test the nearby, I
just switched on/off enable nearby!

Agree, transpose (and everything else) is working better (may find bugs
later, but that's part of the fun!).  I'll try to build later for the store
- thanks again for your considerable input!

I've biy the bullet and just ordered a MacBook Air M1 for Android
development.  I currently use an old Packard bell machine running Manjaro
Linux, but it's 11 years old and a single core device!  I updated the ram
to 8Gb and put in a SSD drive about 2 years ago, which helped, but it's
really slow now.  It takes about 12 mins to initially build the app to test
on a device and have up waiting for the emulator to start after 60 mins!
What are you using out of interest?

iv-gha:
Hello Gareth,

Ah, I was testing poor Wi-Fi / wandering out of Wi-Fi range with my 'Faraday' cake tins - the App handles this external change well - silently reconnecting when back in range.

I use a Laptop PC, Intel I5 Dual Core, 8Gb and SSD with large monitors.  I am getting compile times under a minute.  You deserve a medal for endurance!  Your MacBook should fly - joy.

The removal of comments from Lyrics when displaying 'Lyrics only' may need mention in the notes, songs that use round brackets around actual lyrics will need minor rework.  I had a few where the women part was in brackets, but many more with comments that should not be displayed (got to love SongSelect!).

We have put together some good stuff for release.  Hopefully it is not too much work to publish on the stores.

Kind regards
Ian V

iv-gha:
Hello Gareth,

When you clear the connection log is there a way of adding a list of currently connected clients?  It might be a way to give a list of who is currently connected.  Cannot see a nice nearby call - doh!

Kind regards
Ian V

thebigg73:
Similar to the remembered clients, we could store them in an array, but
remove them on disconnect/lost?

iv-gha:
Hello Gareth,

In testing, if a client is connected then a change of 'Act as Host' does not work.  I previously got irritated with unstable switch of mode and took the approach of clicking nearby off and in the code (as manually this is what fixes). I realise this is a kill and start of nearby - it is however reliable and so I recommend it  as a full restart refreshes nearby - removing any doubt as to the state.

Kind regards
Ian

thebigg73:
That isn't the case.

Google Nearby actually doesn't care who is the host once a connection is
made.  This is based on the P2P connection mode.  You do need a host to
initiate the connection, but once connected, all devices are seen as
equal.  They can all send and receive.  I use that isHost variable to
decide if song details should be sent to listening clients when a new song
is loaded.  When listening, the app checks that it isn't a host before
responding to incoming payloads.  However both clients and host can send
and receive payloads, in just filtering out the appropriate ones in the
song load and payload received

iv-gha:
Hello Gareth,

Sounds good.  

So, nothing active.  Set 'Act as Host',  wait for connection, confirm to connect to client.  Unset 'Act as Host'.  Device does not enter Discovery as P2P remains active.  I think this needs a log entry 'Connected to other devices. Waiting for an 'Act as Host' device to send messages.'' or something like that to give the users info?

When in this state they can return to 'Act As Host' , it does not log that it is broadcasting.  If this too is expected then perhaps log the return to broadcasting?

At the moment we just say stuff stops.

Kind regards
Ian V

thebigg73:
Sorry, I realised I never gave the full story...

When someone acts as a host they stop discovery (if they were doing that as
a client( and start advertising as a host.  Any existing connections remain.
If someone was a host but then switched host off, they stop advertising and
start discovering, but any existing connections remain.
The log should reflect those changes though already

iv-gha:
Hello Garerh,

I observe, Changing 'Act as Host' when no connections does log discovery / Broadcast  changes as you have detailed.

Test this,   Set 'Act as Host' on Device 1, connect to Host Device 1 from Device 2.  Change 'Device 1' - unset 'Act as Host', observe no Discovery,  set 'Act as Host', observe no Broadcast.  Does this happen for you?

I observe, changing 'Act as Host' when connected does not log discovery / broadcast change as you have detailed.

Kind regards
Ian

thebigg73:
Yeah, I think you're right.  I think the code will only do the switching
off discovery/advertising if no connection is currently made (that was to
stop people switching on act as host after nearby).  I'll adjust the code,
but definitely no disconnect/reconnect process should happen.

thebigg73 (changes requested):
Hi,

I've made a fair few changes to the way Nearby is presented to the user.  As a result, I think this commit is now invalid.  Can you check

iv-gha:
Hello Gareth,

Sounds interesting - I will certainly take a look.  A fair few changes may
take a fair time !  Good stuff.

Kind regards
Ian

iv-gha:
Hello Gareth,

I have looked at the improvements and made a few layout tweaks to hopefully help users to transition.  See what you think. I try to adopt the general menu style.

The approach you have taken is clearer - great.

Kind regards
Ian

thebigg73 (approved):
Looks good
I'll merge and test

PR #107, opened 2021-03-26
Nearby and HDMI/Second screen adjustments 

iv-gha:
Hello Gareth,

Glad to hear your Mac is flying!

I needed a 'Lyrics Only' format for 'Stage Mode' and took the opportunity to fix HDMI and mode change quirks.  These give 'Stage Mode' a new 'Lyrics Only' format (and as a bonus greater alignment flexibility for 'Chords and Lyrics' format in presentation mode).

A user in stage mode with HDMI output does not have to connect to another device to get 'Lyrics Only' displayed using that device.  They might be happy to use Stage Mode in this new Lyrics Only format and cable up their own device!

I have tweaked nearby to allow users to change their mind about settings or simply re-display the last host song.

I would love some feedback on these changes.  I am happy to adjust.

Kind regards
Ian V

### DIFF
--- a/app/src/main/java/com/garethevans/church/opensongtablet/FullscreenActivity.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/FullscreenActivity.java
@@ -92,6 +92,10 @@ public class FullscreenActivity extends AppCompatActivity {
     public static int whichPad = 0;
 
     public static boolean alreadyloading = false;
+    // IV - Some mode oncreate actions are only needed once
+    public static boolean doonetimeactions = true;
+    // IV - Persist knowledge of a current hdmi session
+    public static PresentationServiceHDMI hdmi;
 
     // Flag to indicate that Android Beam is available
     public static final boolean mAndroidBeamAvailable  = false;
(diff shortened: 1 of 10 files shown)


thebigg73 (commented):
Looks good on skimming through - I'll test!
Thanks Ian

iv-gha:
Just noticed align top and bottom not working in stagemode 'Lyrics only' mode (an '|| stagelyricsonly' is needed somewhere) and that in presentation mode when font is not autoscale behavior is not right with block text background going full width - this last one feels like an old bug.

Added:
For the stage mode align this 1 line change will fix
ProcesSong.java
2851         if (StaticVariables.whichMode.equals("Presentation") || (StaticVariables.whichMode.equals("Stage") && !(preferences.getMyPreferenceBoolean(c, "presoShowChords", false)))) {

I cannot find where the presentation problem is - in presentation if you turn off autoscale lines align left and are full width (using block text shows these are full width lines).  Align left/centre/right  is probably working but on a full width line has no effect.  Perhaps you can fix?

iv-gha:
I have push a few further tweaks to my branch (includes the fix mentioned).  Mostly to tidy up layout when using top and bottom alignment.

"I cannot find where the presentation problem is - in presentation if you turn off autoscale lines align left and are full width (using block text shows these are full width lines). Align left/centre/right is probably working but on a full width line has no effect. Perhaps you can fix?" - is still a problem.

Cheers
Ian

thebigg73:
Holidays start after tomorrow!  That means I'll have a bit of time to catch
up with everything OpenSongApp!  I've been totally timed out recently (a
school teacher by day and it's been hectic!).
Thanks again Ian.

iv-gha:
Hello Gareth,

I am thinking that copyright  CCLI info option should be available for all projections.  I am thinking we use the presentation info block all the time which gives the user size and position options. Less code too!  How does that sound?  I have had a go and it looks good.

Kind regards
Ian

Edit: See #108!

PR #108, opened 2021-04-02
Connected Display Tweak (2)

iv-gha:
Hello Gareth,

This moves to using the 'Presentation' presentation info for all modes.  A positive (I think) side effect is that backgrounds can now be used in all modes.

Further options for presentation info we might consider are:
1. Background colour opacity (0 being OFF)
2. Autohide ON/OFF

These would give a user the ability to re-establish the old song info. No background colour, autohide OFF and set copyright info size to 0. No logo though!

Please consider.

Kind regards
Ian V

### DIFF
--- a/app/src/main/java/com/garethevans/church/opensongtablet/ExportPreparer.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/ExportPreparer.java
@@ -211,7 +211,6 @@ class ExportPreparer {
         Uri onsong = null;
         Uri image = null;
         Uri pdf = null;
-        // IV - Moved pdfbmp into relevant PDF section
 
         // Prepare the song uri and input stream
         Uri uriinput = storageAccess.getUriForItem(c, preferences, "Songs", StaticVariables.whichSongFolder,
--- a/app/src/main/java/com/garethevans/church/opensongtablet/PopUpLayoutFragment.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/PopUpLayoutFragment.java
@@ -257,7 +257,8 @@ public class PopUpLayoutFragment extends DialogFragment {
             setCheckBoxes();
         }
         setXMarginProgressBar.setMax(150);
-        setYMarginProgressBar.setMax(150);
+        // IV - Allow bigger margin to give smaller active height suitable for use as bottom 3rd overlay of words over video
+        setYMarginProgressBar.setMax(250);
         setXMarginProgressBar.setProgress(preferences.getMyPreferenceInt(getContext(),"presoXMargin",20));
         setYMarginProgressBar.setProgress(preferences.getMyPreferenceInt(getContext(),"presoYMargin",10));
         setRotationProgressBar.setMax(3);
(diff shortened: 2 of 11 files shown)


thebigg73 (approved):
Pulling in your changes

thebigg73:
My word - what a cracking job on the presentation windows!  You've even got the chords lining up with the lyrics when set to center horizontal and I'm impressed you've overcome (using transparency) the blocks behind song sections in performance/stage modes to allow second screens to work with backgrounds.  I'm delighted!!

I've tested all of the changes I can see and they're all working fine.
I have a bug fix for KitKat users not able to select storage (due to some of our recent changes to the BootUpCheck file that we only focused on fixing names for document tree uris and non uri files were getting ignored and therefore disabled).  I'm keep to push as v5.3.2 if you're happy.

Job well done again Ian!

iv-gha:
Hello Gareth,

Can you figure out why using manual font size makes it go full width in presentation mode - feels like it is behaving as a chords line which go full width! Odd. Not quite right there.

Glad you liked the change.  It would be good to have the opacity and auto  on/off to give max flexibility for those chroma keying in the info block (ccli) over video.

Good you have a fix for location! Whoops. Can I suggest that use of yellow find button opens the info block.  If they are doing a find the info might help.  Also I suggest we change the icon to a question mark - so it is a bit more overtly  'help'ish?

Lots of driving today so I will not be able to respond! Have fun.

Kind regards
Ian V

thebigg73:
Enjoy your driving!
I'm trying to remember what I did with the autoscale off, but if I remember
rightly I tried to get it to automatically wrap the lyrics across multiple
lines but obviously this messed up chords, so they had to be off.  The text
box was set to full width regardless to allow the wrapping.  It's been so
long, so I'll need to check back!
For the info bar and chroma keying, do you mean when using the built in
videos or making it compatible with external video mixers?  I thought the
info bar already had transparency?  If not I could add that pretty easily
(I think) along with toggle switches.

Not expecting a reply today, so no hurry.  I also run a microbrewery in my
spare time and I'm hosting a virtual home brewing session this afternoon
teaching people how to homebrew 1 gallon of beer in their kitchen with a
grain recipe we supply them with.  It's been a great way to trial and test
it new recipes!

I've also been getting on with the material app version.  I'm most of the
way through the settings and menus now and I'm getting close to getting the
songs and features working.  I'm having to manually import your updates
into it as the file structure is radically different and the code is (I
think) far better for performance.  I'm trying my best to keep both going!

Best wishes,
Gareth

iv-gha:
No problem.  I have found the wrapping - and it is what breaks align of lyrics (no chords) when using manual font size, it also shows just the last line if it wraps - doh).  Align is perhaps more important so I have removed the broken wrap and tweaked PopUpLayoutFragment to match.  Like performance mode, a manual font choice runs the risk of overrun and we provide autosize to avoid.

Use for overlay is indeed with external video mixing - useful for lockdown Zoom meets!

I do not believe infobar background has an opacity setting for users.

We did not add an on/off option for auto-off when it was introduced.  For presentation we would need to make sure Alerts still work whatever the setting.

Can you add the options - I can then add my bits.

The pace of tweaking is perhaps slowing now.

Having fun with Beer too!

Kind regards
Ian V

thebigg73:
Hi Ian,

I think I've fixed the alignment thing separately.  If chords are off (and autoscale is off), the lyrics will wrap across lines (this was requested by quite a few people so I want to keep this in).  I've fixed the alignments of the wrapped lines too.  If chords are on (and autoscale off), then I've stopped the wrapping, but kept the alignment.  The user will run the risk of running out of space, but that was their choice in switching off autoscale!

I'll look at the options for opacity and autohide now.

thebigg73:
Hi Ian,
That should be the opacity and autohide option for the info bar now working.

Let me know if I've missed anything else we talked about!

PR #112, opened 2021-04-17
Enhancements for presentation lyrics wrapping / info bar hiding / Nearby song send

iv-gha:
Hello Gareth,

I think I have fixed wrapping lyrics - take a look.  It took quite a while to find the few lines that needed changing! 

You get a few code logic simplifications made as I went along trying to understand.

It also contains my take on supporting the new options - I have the advantage of having designed the hide song info logic.

For consideration.

Kind regards
Ian V

### DIFF
--- a/app/src/main/java/com/garethevans/church/opensongtablet/BootUpCheck.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/BootUpCheck.java
@@ -849,6 +849,7 @@ public class BootUpCheck extends AppCompatActivity {
     }
 
     private void startSearch() {
+        MyQuicktip.setVisibility(View.VISIBLE);
         // Deactivate the stuff we shouldn't click on while it is being prepared
         setEnabledOrDisabled(false);
 
--- a/app/src/main/java/com/garethevans/church/opensongtablet/BootUpCheck.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/BootUpCheck.java
@@ -860,6 +860,7 @@ public class BootUpCheck extends AppCompatActivity {
     }
 
     private void startSearch() {
+        MyQuicktip.setVisibility(View.VISIBLE);
         // Deactivate the stuff we shouldn't click on while it is being prepared
         setEnabledOrDisabled(false);
 
--- a/app/src/main/java/com/garethevans/church/opensongtablet/NearbyConnections.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/NearbyConnections.java
@@ -55,6 +55,10 @@ public class NearbyConnections implements NearbyInterface {
     private final ArrayList<String> connectedEndPointsNames;
     private final ArrayList<String> connectedDeviceIds;
 
+    // Handler for stop of discovery
+    private final Handler stopDiscoveryHandler = new Handler();
+    private final Runnable stopDiscoveryRunnable = () -> stopDiscovery();
+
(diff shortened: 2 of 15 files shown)


iv-gha:
Hello Gareth,

I had to address for non AirTurn mode onLongPress not firing after use of a Fragment (Android 5, was not a problem on Android 9) - for example selecting a song from a different song folder.  If the next pedal action is  a long press, onLongPress does not fire!  A 'panic' has been added to long press based on keyRepeatCount.

For AirTurn mode - I use long press to change sections of songs (less obvious than stamping a short press!).  I step and hold, release, step and hold.  When testing AirTurn mode this was not reliable as the code had a long 400ms minimum release period.  I have shortened to 100ms.  IT NEEDS TO BE LONGER THAN THE KEY REPEAT PERIOD - IS THIS PERIOD LONG ENOUGH?  I am wary of breaking things here!  It should be quicker than the human can release and re-press the pedal.

I add a 'only one action per pedal use' mechanism that supports the above approaches.

Hopefully useful.

Kind regards
Ian V

thebigg73:
Hi Ian,
I'll pull these in.  Regarding the AirTurn mode, I don't have an AirTurn pedal for testing, so the code has always been my best guess of trying to get it to work and I think it has limited success anyway.  From your notes, it sounds like you have one and if it is working for you, then I'm more than happy to trust your coding around it.  The key release period 'should' be ok, but again, without an AirTurn pedal, I'm just guessing!  We could always leave it shorter at 100ms and if we get complaints, we could up it a little?

iv-gha:
Hello Gareth,

The changes in https://github.com/thebigg73/OpenSongTablet/pull/112/commits/c00ce8dbf84e5dca270564a5fc836319be6823db
are the 'Nearby Song Send' are the other area of change.

On reflection the test for AirTurn boils down to is this onKeyUp the last one in a run and to know this we test after a 'repeat delay' time.  Looking at the scant info I have found on AirTurns I have seen this as 610ms (on a settings screenshot) and as 2s(!) for the BT105 (in manual).  This 'repeat period' is important as a short press cannot be fully detected until we know no key event occurs at the 'repeat period' - we delay the short press by the duration of the postDelay 'repeat time + a bit' period!  You had 400ms before (200ms + 200ms) but I think 620ms would at least be after the 610ms (settings screenshot).  2010ms would be unusable - a short press with a 2s delay!

We could ask AirTurn what the standard timings are for their pedals, time to first key, repeat period, timings of subsequent key repeats.

My repeat period test at 100ms is certainly not good.

I do not have an AirTurn but may be able to borrow a friends (may not be for a while though).

As interim it could be set to 400ms as before.

Thinking....

Kind regards
Ian V

ADDED:  I think that my use of long press to change songs only works well with pedals (like mine) that support 'OnLongPress' event.  AirTurns approach is bad!

thebigg73:
Yeah, my homemade pedal works fine with onLongPress too.  I've only ever had one proper request about the simulated long press for users of AirTurn pedals - I'll need to see if I can dig that out!  

Would it be worthwhile to give a user seeing with a seekbar to let them change the repeat period?  There is an SDK I can download for AirTurn too I think (I've communicated with them several times, but they're not great at getting back and when they do they insist their pedal works in the same way as all others).  The SDK might give more insight.  I'll let you know.

iv-gha:
As the airturn user can alter the repeat period...  A user set-able repeat period gives best flexibility. This would cope with all pedals that repeat using up/down pairs after a repeat period. Great idea.

PR #114, opened 2021-06-04
Stage Mode: Clickable area for page buttons increased.

iv-gha:
Aim: To reduced the frustration of button touch fails, particularly long press, as the user can now be on or near the button for action to occur.
A button is placed in a RelativeLayout with clicks of this layout passed to button click routines.
NearbyConnections: Minor change.

Hello Gareth,

Please consider this tweak which is my attempt at making it easier to use buttons.

Kind regards
Ian


thebigg73 (approved):
Look sensible.  I've updated the Material App to be more logical with page buttons, so hopefully a redundancy fix for this current version.

thebigg73:
I've pushed this fix along with the others to the beta channel on Google Play.  I had to tweak a few storage permission errors earlier to allow the uploading (Google Play changed settings that stopped me uploading updates until I made some changes).  Since this makes me nervous for KitKat users (really struggle to test on KitKat nowadays), I've pushed to Beta rollout (affects about 400 users who agree to potential issues) rather than the stable channel.  If there are no major issues, I'll then push it on to the stable channel for everyone else.

Many thanks again for all of your hard work submitting fixes and updates!

PR #116, opened 2021-07-05
Non DB song renames/duplicates. Selecting variations. Song list checkbox handling.

iv-gha:
Hello Gareth,

This makes rename/duplicate of received and variation songs work better.   A received song can be duplicated - this was possible but a bit broken.  Variations can be duplicated - a song passed as part of a set can be saved.

I was trying to address rehearsal pain -  I use receive host songs and want to retain only the host songs I choose (so no 'keep host songs' as this disturbs my version of songs!).  I use a Temp folder.  For a song I want to keep I will duplicate to the Temp folder and use and adjust this retained song as needed (and return to later in rehearsal if the host sends their version again).

After rehearsal I can calmly consider what to do with the Temp songs - save as variations in a set and delete (store in a set), move to a different folder (add to songs) or discard.

EDIT: And other stuff!  I looked at song checkboxes and have added a layout and click action (none) for the area around them. This, I think, stops a recently reported issues of unintended list movement if this area is clicked.

Kind regards
Ian


thebigg73 (commented):
I've pulled these and merged the appropriate ones into the MaterialApp branch.
There is one bit of code that needs changed (comparator & to change to && in the PopUpSongDetails fragment title).  I couldn't do this before merging, so rather than bouncing the whole thing back, I'll just change it at my end.

iv-gha:
Good spot.  I have put another pull for a song list problem.

Thanks for taking a look.

Kind regards
Ian

iv-gha:
For fix
"This handles clicks near the checkbox, preventing unintended movement of the list (maybe!) - now nothing happens."

Just for info - Android reserves long clicks near to the (hidden) scroll bar for control of the scroll bar.  So a long click in that area only goes to the scroll bar and will scroll the song list to the selected position.

PR #119, opened 2021-07-28
FIxes

iv-gha:
Hello Gareth,

Some fixes and features.

I have adjusted PrepareSongMenu as the code intended to remove entries for deleted songs but was broken.  When broken the song menu opens quickly as checks are not done.  Now it is fixed the 'Wait...' whilst it checks files occurs.  The larger the number of files the longer time.  Users might notice!

I am interested in your impression of the take up of use of modes.  Of the 13,000(?) active users how many are presentation or stage mode users.  Is the number of active users going up?

Please consider.

Kind regards
Ian

### DIFF
--- a/app/src/main/java/com/garethevans/church/opensongtablet/ChordProConvert.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/ChordProConvert.java
@@ -935,8 +935,8 @@ class ChordProConvert {
                     if (thisLine.startsWith(".")) {
                         thisLine = thisLine.replaceFirst("."," ");
                     }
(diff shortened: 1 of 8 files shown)


thebigg73 (approved):
Should be fine.  Some of these will be rewritten using Java objects rather than just using StaticVariables (this file along with FullscreenActivity will be removed).
e.g. Pads/mediaplayers will be objects that handle their own actions rather than being stored as they currently are.   In the same way that songs are now objects rather than just a bunch of variables.

thebigg73 (approved):

PR #120, opened 2021-08-02
Song layout update and highlighting improvements.

iv-gha:
Hello Gareth,

I have managed to take a first look at MaterialApp and the interface changes.

I saw some song chord highlighting issues and a need to bring forward some of the Master layout improvements - I have had a stab which is offered to you.

The render speed is faster - what has brought about the speed improvement?

Kind regards
Ian


thebigg73:
Hi Ian,

D you can see it is still far from being ready!!  The song display isn't
right and only uses simple preceding just now - I just needed enough to be
able to see it in the display.
The metronome is working nicely now I think with every tune signature
likely to be needed including complex and compound timings.   It doesn't
save to the song yet.
Autoscroll, stage mode,  presentation mode,  casting, pads are non existent
for now and most features are still work in progress with known bugs.
Still, you can see how I'm seeing out the app.  I wouldn't worry about
trying to fix stuff unless you see anything obvious!  I am trying to
include the changes from master that are appropriate to the new app
structure (some will be obsolete), but I probably have missed some.   I
know there is lots in processsong that is still missing as I haven't got to
that yet.

One feature i think you'll like (i think i left it in a working state) is
the export song.   You can choose a pdf and the app will now render a nice
pdf version of the song with non monospace font.   It makes use of the page
drawing method described below that is also responsible for the improved
render speed.

The render speed improvement is down to how scaling is figured out.   In
the current app, the layouts are rendered for 1,2 and 3 columns, measured,
checked for size then the best size/ required scaling method is drawn to
the screen by scaling the font size.
Basically the song is processed and drawn 4 times.
Now I'm drawing each section once to a test layout, measuring them,
detaching them from the test layout then reattaching them to the display
with the view scaled by zooming in/out using scale x/y  The processing only
happens once now and will be reusable for the 3 modes.
Another bonus is the zoom better fits the screen rather than relying on
scaling the font size (which sometimes overshot or under shot the screen
margins due to rounding).
Because I have skipped some song processing methods for now, there might be
a slight slow down once i put them back in.

iv-gha:
Hello Gareth,

Sounds good.

The pull fixes the new chord highlighting and adds the current layout approach so you should find it useful.

I will have a look at the new features.

Thanks
Ian

iv-gha:
Hello Gareth,

So interesting... I noticed that the song menu did not have the file path present on the original search which I use - this set me looking at the code and I though you might like to see the outcome.  You will see minor layout changes too - intended to help usability.

The user can get something similar to the original song menu by having only the folder search on.  It would not take much to give an option to sort by filename (original) or title (new) - this is just a change to the order clause of the revised SQL.

I hope you find something useful.

Kind regards
Ian

iv-gha:
Just a note: A fresh install of the the MaterialApp version is not launching.  I cleared data on mine and it would not restart - problems with storage selection. Doh!  This is for a fresh 'as-is' version from the MaterialApp branch updated a few minutes back.

Kind regards
Ian

iv-gha:
I'll close this pull (as you have included bits) and rebase at some point from the current state of the branch.

Kind regards
Ian

thebigg73:
I tried to pull it in but it kept trying showing changes that weren't
there!!  I had started the pull in Android studio rather than on the GitHub
website which I normally do, so I'm not sure if I bring something!!  I
manually added everything instead - took a while

Cheers!

Gareth

PR #130, opened 2021-08-21
Bug, crash and feature fixes

iv-gha:
Fix: Song edit (relevant fields only) and note edit support for PDF and Image songs in all modes.
Fix: Correct title and (when edited) capo info display for PDF and Image songs.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -58,7 +58,7 @@
         android:fullBackupContent="true"
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
-        android:largeHeap="false"
+        android:largeHeap="true"
         android:roundIcon="@mipmap/ic_round_launcher"
         android:theme="@style/MyTheme"
         android:vmSafeMode="true"
--- a/app/src/main/java/com/garethevans/church/opensongtablet/AutoScrollFunctions.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/AutoScrollFunctions.java
@@ -203,7 +203,7 @@ class AutoScrollFunctions {
                     tvs.setTextSize(StaticVariables.infoBarSmallTextSize);
                 }
 
-                if (currtimesecs < StaticVariables.autoScrollDelay) {
+                if ((currtimesecs < StaticVariables.autoScrollDelay) && (!(FullscreenActivity.isPDF && FullscreenActivity.pdfPageCurrent > 0))) {
                     // Set the time as a backwards count down
                     currtimesecs = StaticVariables.autoScrollDelay - currtimesecs;
                     text = TimeTools.timeFormatFixer(currtimesecs);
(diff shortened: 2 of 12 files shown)


iv-gha:
Fix: Pedal action and on-screen button actions corrected (current BETA is bugged)
Fix: Scroll button use for zoomed PDF to support page turn

iv-gha:
Fixes: highlightNotes usability and stability when using large images
Feature: Added use of fling for PDF page turns

iv-gha:
Fix: PDF Autoscroll
Fix: No bottom info block for song only with title starting with '_' (which hides the title)

Comments:
We perhaps say in the app 'blurb', 'PDF (Android 5+) and Image files are supported for simple display only'.  Try to head off those wanting a full featured PDF/OCR app!
Add 'Song Book' to the title to indicate a prime use?

thebigg73 (approved):
Initial checks look ok.  I've manually added a few into the materialapp branch, however, I've ignored most for that branch as they are incompatible.  Fine for the master branch though

thebigg73:
For the upcoming (eventually) MaterialApp, I am planning on much more PDF support (full song fields and a built in OCR converter tool which will also try to extract text from photos too).  Song book in description is a good call.


Comments:
We perhaps say in the app 'blurb', 'PDF (Android 5+) and Image files are supported for simple display only'. Try to head off those wanting a full featured PDF/OCR app!
Add 'Song Book' to the title to indicate a prime use?

thebigg73:
Would you consider this ready for replacement of the beta version?  Maybe push to stable channel if ok too?

iv-gha:
Hello Gareth,

Most of the changes made remove code restricting features for PDF and Images and tidy up what is then made available - so  appear safe.  A user could not do these things before.

The HighlightNotes changes are I think quite safe - we hide incorrect display for a zoomed song, but users are aware that full screen is what is needed for highlightnotes.  However if  it is bugged then it affect performance.

Changes to scroll and page buttons (direct or via pedals) need great care as if wrong it affects performance.  I think the change to use of scroll for PDF is good (same behaviour as opensong songs) but I do not use PDF!

You should perhaps test usability to gain further confidence - making sure use in performance is good across all song types. (full screen+  highlightNotes and autoscaled).  I'd much rather you uncovered any problems.

If you are then happy then I think BETA for a while.

Kind regards
Ian

thebigg73:
I've had a quick play and agree the swipe up/down for PDF is quite a neat
way of dealing with the songs.  My pedal is currently out of action, so I
can't check that.  Looking at the code logic, however, it seems right -
scroll (either as a default action or scroll before next/prev) moves
through the pages, but next /prev will move to the next/previous song.  As
long as I include that in the release notes, its fine - as you say,
OpenSong format is preferred anyway and PDF support is a bonus.

The restrictions for PDF were not intentional!  Similar to you, I don't
really use them so often forget to test them properly.  The Material app is
planned to give all the same features other than edit/transpose, but will
give the opportunity via OCR to convert to an OpenSong formatted song.  If
users insist on PDF, they can still have highlights, pads, metronomes,
links, midi, etc but just can't edit the song display (a real pain in my
opinion, but expected for PDF).

I'll check the scaling/highlighter stuff later (I'll need to push to my
phone for testing as it's not easy zooming on an emulator!).

As long as there are no show stopping crashes to the best of our knowledge,
beta is a perfect place for testing to iron out the minor bugs.  I've got a
busy week/weekend, but will try to get this tested and uploaded to the beta
channel soon.

As always, many thanks for your contributions and time you spend on this!

Gareth

PR #133, opened 2021-09-22
Major update to Nearby section change handling

iv-gha:
Hello Gareth,

A host sending section changes and a client using section changes are kept in better step - important for 'Stage' controlling 'Presentation'.

I did not want to throw out first and last song (with delay) being sent to clients during rapid song changes  This handles section changes in the delay period - previously changing the clients old song and not being applied to the new song!  Sorted.

I think I need a rest!

Kind regards
Ian

### DIFF
--- a/app/src/main/java/com/garethevans/church/opensongtablet/LoadXML.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/LoadXML.java
@@ -326,7 +326,7 @@ public class LoadXML extends Activity {
         StaticVariables.thisSongScale = preferences.getMyPreferenceString(c,"songAutoScale","W");
 
         // IV - When an error force use of section 0
-        if (FullscreenActivity.myLyrics == c.getString(R.string.user_guide_lyrics)) {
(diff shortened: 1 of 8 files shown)


thebigg73:
You have been busy!!  I'll get a look shortly - work at the day job has
been chaotic just now and I've been struggling to make time to get a
serious look at your (likely excellent) work.  Trust me, I really
appreciate your input with this!
Gareth

thebigg73 (approved):
Pulled in!  Look good - I've not manuallly copied in the stage mode sections work yet to MaterialApp as that is currently non-existent!  I'll need to try to remember!

iv-gha:
Great. I think I have found all of the rabbit holes that PDF and Image
handling go to!  Have fun.

iv-gha:
Hello Gareth,  just a note about nearby...  If 'receive host songs' is off on a client then only BYTE info is used and the FILE transfers are cancelled.  This makes all song types fast on the client, all use BYTE and local songs.  Before this change songs with FILE would not be progressed until the (unused) FILE completed.  A client will also cancel older transfers when a new song arrives.  Previously old FILE transfer continued resulting in an unneeded song being displayed.  Cancel is very good!

PR #134, opened 2021-10-03
New Nearby options and better loadsong behaviour

iv-gha:
Hello Gareth,

New options for autoscroll and section changes with Nearby - to give the client user control.
loadsong error handling changed to use opensong XML for errror reporting.

5.3.7 is so old hat!

Kind regards
Ian


thebigg73 (approved):
Looks good.  Pulled appropriate stuff into MaterialApp too.
I think it is time to update the current stable version!

PR #135, opened 2021-10-20
Fix: pad logic moved to animateInSong

iv-gha:
More...
Running pads and metronome was significantly slowing song display .  Rapid song change was causing multiple pad starts.  Latest commits try to remove these performance pains!

Kind regards
Ian


iv-gha:
During performance this guarantees a 2s grace period after full song display before a pad change.  Change song again and the original pad keeps playing until settled on a song for 2s.  Great for capo changes, 2s to get set before the pad change - hopefully less perceived delay to the start of the next song.

Yes, I think there is a good weight of performance and feature improvements to release.

thebigg73 (approved):

PR #136, opened 2021-12-01
Perf: Transpose and song section preparation

iv-gha:
Hello Gareth,

This commit gives a a real performance boost!

This is mostly about Transpose and section prep which I looked at because of slow songs changes when performing.  As you are looking to use Transpose with set songs this is, I believe, useful - for Material it should fit with the view work you have done already.  Fast enough to slow down!

Please read the commit notes for details - happy to discuss.
 
You also get a few fixes and features addressing pain points - like quickly adding a variation and changing its' key.

Christmas present!

Kind regards
Ian V

p.s. This is a large commit - apologies for the work integrating into Material!


iv-gha:
Hello Gareth,

Notes:
1. There are still some information Log.d in the code.
2. The language strings will need updating if the chords dialog changes are good
3. I re-based against your master prior to this commit and it threw up build errors because Cast went up a version.  So please check cast still works!

iv-gha:
Do wide variations in transition times give a poor user experience (UX)?  I came to the conclusion a quick out animation to start render quicker, a minimum render time to reduce variations in overall time and a 300ms animate in for effect.   I have tried other combinations - if we have animations then giving the time to animate in seems the best UX.  Now similar UX on all devices.

iv-gha:
I hope the term has ended well for you and you are able to recuperate!  Enjoy Christmas.

I think there is probably enough for a fresh release in this commit.  I added long press across the menu bar to help with common activities (at least common ones for me!).

When helping a friend I noticed the Amazon version is at 5.3.2,  so perhaps give those users some love this time round.

Have fun.
Ian

thebigg73 (approved):
Lots of change here!!
I've pulled some obvious stuff into MaterialApp, but there is a lot of redundant stuff that I've ignored for now (as it is no longer required, or will require me to rewrite to match newer render methods for song sections).  I'll try to test over the next week or so.
Thanks again, and hope you have a nice Christmas!

thebigg73 (commented):
I tried the chord style conversion and really impressed with Nashville etc.  I did notice that when converting to European mode (ie. Bb=B and B=H), when transposing this format it messed stuff up.  i.e. B (which should have been Bb) when +1 transpose should go to H, but it went to C.
I'll need to dig deeper into your new transpose method!

iv-gha:
I think we need in and out format control on transpose( and use preferred).  The in would show (bold) the detected but pre select the assumed if set.  The out the same.  User will see what is happening and have control - at the moment it is a bit hidden.  We should later think about adding an undo as well!

I agree German transpose looks broken - let me look.  I fixed lots of others - hadn't spotted this one as it is close to standard - and there lies a problem , close enough to mis-detect!

iv-gha:
Improved German detect line as below that finds more H...
The commit adds mis-detect thresholds which require more than 2 H#s to be detected.

Transpose,java
554                } else if (line.contains(" H") || line.contains(".H") || line.contains("/H")) {

A bug from the older code looking for \ is now found!

If we go for in/out choice for transpose then the same approach (no slider) will work for Use preferred.

thebigg73:
Cheers Ian - yes maybe a transpose 'from' and 'to' rather than detected?
Currently looking at this in Material App
image
image

iv-gha:
Thinking of detect - explore this new feature... convert some variations to German, some to Solfege, have some as Standard.  Set in choose chord format for 'Choose your preferred...' Nashville and set 'Always assume...' off.

View your variations - the source is German/Solfege/Standard but all display in Nashville.  In is using detect and out is fixed to Nashville.   The chords are Capo colour as none are native (Capo 0 when no capo).

Show both Capo and Native chords and you see Nashville above the native chords.

A further case.  A Host has a German song and views in Nashville (in detect / out fixed Nashville).  A client receives the German song and displays in Solfege (in detect / out fixed Solfege).  Neither have converted the song.  Both Host and Client choose what they see.

What is the app default for 'Assume..'?  

The trick is getting dialogs that make what is going on clear including the above feature.

Certainly having in and out control in Transpose and Use preferred is good.  I did work in visibility of what is detected (using bold - perhaps not obvious enough) to the commit.

I am letting my brain get happy with these new features as I am wating for a 'oh, that's how to do it' moment!  The Transpose.java approach in the commit was after many 'oh' moments.  The good stuff consolidated and optimised.

We need detect for the above feature - this makes coming up with good dialog design harder as it needs to include detect info!

I think to release this transpose code as it was not working correctly previously for Solfege etc.  Dialog change needs more care than typical.

Kind regards
Ian

thebigg73:
I've done some testing with the transpose and have come across some minor bugs.  I've fixed in MaterialApp and I'll try to map them to the master branch to match.  Involves changing A# to B in format 2/3 (matches Bb), resetting the Nash strings[] on transpose call as before it would only allow one conversion (as it looked for null values before it worked), also there was a check for the newChordFormat when it should've been oldChordFormat.  I've also improved the chord format detection - your method only counted lines, whereas I've now got it counting actual chords.  Seems to be more reliable as I noticed a song with only 4 chord lines was struggling to be detected as anything other than standard.

I've also updated the new transpose window in MaterialApp.  I don't think it is worth trying to backport this as it might be too much effort given that master will hopefully soon be deprecated - I'm getting really close to MaterialApp being an alpha release.  The main bits still to do are the PresenterMode, Casting and Stage Mode sections.  Performance Mode is pretty much done with all associated features.

Here's my plan for the Transpose popup in MaterialApp showing not using preferred and when preferred is on, but detected is different:
image
image

I've managed to convert back and forward between all formats several times and the chords convert/transpose well - excellent job on figuring out this simpler transpose method!!

Gareth

thebigg73:
I think I've updated the required code to deal with my changes in MaterialApp regarding the bugs/format check.  I've also identified the issue with some chord formats showing as capo chords

thebigg73:
Hi Ian,

I've uploaded to the Play Store and Amazon Store (it was sitting with a
draft version of 5.4.1 which I now remember was during a 2 day outage on
their developer console and I couldn't complete the upload - then I
obviously forgot!).  Both now have 5.4.2 (although Amazon will probably not
be live until tomorrow).
The convert works nicely using your new transpose method on MaterialApp and
on the Master branch, albeit with slightly less functionality for now.  I
fixed the German transpose and improved the format detection too.  I
put a couple of screenshots on the closed pull request comments, but not
sure if you get emailed about closed commits.
Have a great Christmas!!!

Best wishes,
Gareth

PR #138, opened 2021-12-28
checkChordFormat - Improvement especially for format 3 detection

iv-gha:
Fix: StageMode; Stop crash if pedal move to previous (non existent) song in set for a received song.
Opt: Transpose. Minor simplification caused by change of handling leading full stop of chord lines.
Fix: Transpose. Format 3 detection fix. Older checkChordFormat logic had necessary consideration of lowercase chords for format 3.  Worked into the newer approach.

### DIFF
--- a/app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/StageMode.java
@@ -7718,9 +7718,13 @@ public class StageMode extends AppCompatActivity implements
                         } catch (Exception e) {
                             e.printStackTrace();
                         }
-
-                        if (!StaticVariables.songfilename.equals(filenamesSongsInFolder.get(FullscreenActivity.previousSongIndex)) && !isfolder) {
-                            pedalPreviousAndNextConfirm();
+                        // IV - 'Try' fix for crash
+                        try {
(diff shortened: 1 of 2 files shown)


iv-gha:
Hello Gareth,

Congrats on the release - it delivers great render speed improvement.  Great spot on the German transpose errors - I had spotted others carried through from the older code but not those - your fresh eyes caught it.

My eyes have spotted that the new detection format has lost a nuance of the old code in the re-write, I took a first detailed look at the older code, and can see that format 3 needs to test for lowercase chords.  I have added this into the new code.  I also improve chord pre-processing to simplify the chord to better suit the new 'contains' test method.   Detection should be further improved (especially for format 3).  Take a look.

I think this next increment of the check code needs release - you may spot further things I have not.

Your addition of checkchordformat in song section processing is hiding the new 'always show in preferred format' feature.  I am thinking that this could be a preference option in the chords dialog (set off by default) - alongside show capo chords etc.  I am still thinking about how this feature would work / if it is too confusing!

Kind regards
Ian

thebigg73 (approved):
Changes look good.
I don't think it is worth worrying too much about the chord detect/transform idea for the master branch - it might be a little too much in terms of preferences and what to display to the user to make sense of what is happening.  I genuinely think the method that is coming in MaterialApp is far more logical.
I've pulled your changes into MaterialApp as well.
Thanks as always!

PR #139, opened 2022-01-02
Fix: Presentation Mode - high severity fixes!

iv-gha:
Hello Gareth,  fixes for

Serious fault: Cannot edit presentation order in presentation mode.
Serious fault: 1st char of scripture / lyric lines incorrectly removed when show chords is set.
Crash: 1 line section causes crash.

These affect core behaviours of presentation mode - so need to be live soon (before causing complaints!)

### DIFF
--- a/app/src/main/java/com/garethevans/church/opensongtablet/Bible.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/Bible.java
@@ -303,7 +303,8 @@ class Bible {
                 lines = lines.replace("_nl_","\n");
                 newimprovedscripture.add(lines);
                 if (words.startsWith(";") || words.startsWith("|") || words.startsWith("[") || words.startsWith("]")) {
-                    words = " " + words;
+                    // Protect with non breaking space (does not trim) to force display as lyric line
+                    words = "\u00A0" + words;
                 }
                 currentline = new StringBuilder(words);
             }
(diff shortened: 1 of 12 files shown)


iv-gha:
Hello Gareth,

The last commit has sadly directly introduced a much more visible bug - the inability to edit presentation order in presentation mode!  I have added a missing build of the song section list in ProcessSong.prepareSongSections.  I've added for all modes . 
 Other modes got the list as part of displaying all sections.  I fix an older bug too - The commit stops blank section labels from being added to the Song Sections list.

Kind regards
Ian

iv-gha:
Presentation mode in 5.4.2 is stripping first characters - I believe I have found the missing logic and fixed with this commit?

thebigg73 (approved):
I've pulled in a few of these changes into MaterialApp, but there are many which either no longer exist in this format or are now obsolete.  We'll need to test for behaviour of these identified issues when that version is ready for alpha testing.  Not pulled in to MaterialApp- Bible, custom slide, load song, view changes.

I've almost finished Stage and Presenter mode, only leaving Nearby connections to implement after that!

iv-gha:
Hello Gareth,

Sounds good. Mostly a need to put presenter mode back in order.  I think we
release quickly to limit the impact on presenter mode users.

At least I understand a little bit more about custom slide behaviour!

Kind regards
Ian

PR #140, opened 2022-01-15
Fix: For Presentation set list bad display...

iv-gha:
Some song add and removes were not updating the presentation current set!  I have tried to make set display reliable.


thebigg73 (approved):
Look good!  Also translated some into MaterialApp as appropriate.

iv-gha:
A few of my pain points addressed.  Thanks for considering.  Hopefully bug free! (smiley, wink)

PR #142, opened 2022-01-25
Alpha index xml revert and move of padding into code

iv-gha:
Hello Gareth,

I noticed the report of Alpha Index looking odd... as I am familiar with this code I have taken a look.  This commit includes a less 'odd' solution to controlling the width of the index.  I could not re-create the small text seen in the report!  This removes the use of negative margins in xml (might be unstable) and (in code) adds minimum sizing etc.  It is difficult to know it will fix the issue but it seems a safer approach.

Plus other bits carried along with it!

Kind regards
Ian


thebigg73:
The alpha index issue actually happened on my device when I installed it.
It was almost as if the size wasn't initialised.  When I changed the size
to a different value (then back again) it seemed to fix the problem.  I'll
take a look at your changes.  Got a couple of other minor changes to add
too!

Thanks again Ian.

thebigg73 (approved):
Pulling and checking!

PR #144, opened 2022-01-28
Fixes Revised Solfege Transpose

iv-gha:
Hello Gareth,

Here is SongSelect Fixed DO!

I saw the conversation about spacing changes on transpose. We are already code to prevent transpose running chords together - there is always a space between chords.

For your example
```
.G C  D  G
 A short line
```
Transposing this by +3 (to Bb) would give
```
.Bb Eb F Bb
 A short line
```
Chords have wandered - the Eb is over 'h' and F is now over 't'.

For SongSelect (which spaces chords well) the line would start as

```
.G  C     D  G
 A  sho - rt line
```
And transpose to

```
.Bb Eb    F  Bb
 A  sho - rt line
```
All chords are in their correct position.  Reverse transpose would generate the original line.

SongSelect uses a minimum of 2 spaces between chords and ' - ' wherever a chord splits a word.  This means Capo display keeps very good chord positions and transpose works well with minimal need to correct chord positions / adjust to SongSelect spacing using edit.

A well thought out song layout style (thanks SongSelect) avoids much pain  Happily I source songs from SongSelect!

Anyway... enjoy SongSelect Fixed DO!

Kind regards
Ian

### DIFF
--- a/app/src/main/java/com/garethevans/church/opensongtablet/ChordProConvert.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/ChordProConvert.java
@@ -7,6 +7,8 @@ import android.util.Log;
 
 import java.io.OutputStream;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 
 class ChordProConvert {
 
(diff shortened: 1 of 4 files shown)


thebigg73 (approved):
Great work Ian.  I must admit I haven't used SongSelect import for a while, so was unaware of that feature.
I've pulled these changes in and also fixed (I hope) a bug with users getting kicked out of set view.  I've push these updates  to the Play Store as v5.4.5 ASAP as a result

PR #145, opened 2022-02-01
HOTFIX: Song Menu ticks missing after set edit

iv-gha:
Hello Gareth,

Fix for songlist ticks for MAIN folder items disappearing if a set edit is used to change the set.  The root cause is 'doSave' and I have applied the MAIN folder handling to it, this gives healthy set handling / song list ticks (I believe!) - so MAIN is not added in the first place so we can revert to simpler code elsewhere.  Some other bits too.  Please test/consider.

Kind regards
Ian


thebigg73 (approved):
Importing

PR #146, opened 2022-02-07
HOTFIX: Performance mode needs to update song menu ticks after set load

iv-gha:
Hello Gareth,

I am not having a good run.  Apologies but I have caused a further set tick issue(!) which my testing has now found.  With hindsight I should not have pushed other stuff with a root cause fix.  This commit is just a fix for the problem identified and I won't push more to my branch (not add to the pull) until you have collected.

In performance mode, set work updates the song menu (refreshes ticks) on dismiss.

Again my apologies.

Kind regards
Ian

### DIFF
--- a/app/src/main/java/com/garethevans/church/opensongtablet/PopUpSetViewNew.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/PopUpSetViewNew.java
@@ -454,6 +454,8 @@ public class PopUpSetViewNew extends DialogFragment {
 
         void prepareOptionMenu();
 
+        void prepareSongMenu();
+
         void confirmedAction();
 
         void refreshAll();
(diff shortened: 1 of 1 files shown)


thebigg73 (approved):
No worries - Given the number of features in the app, it is so easy to overlook something!  I'm sure MaterialApp will have loads of bugs that you fixed in the master version that I have messed up!

iv-gha:
Hello Gareth,

I see this being missing is causing pain!

In the google group a user is using a set and changes to a new set without
selecting a song.  The song displayed and set buttons do not refresh = same
song with invalid set buttons. This commit refreshes the set buttons = same
song with valid set buttons (maybe none).

As you comment if they select a song from the loaded set then the song and
buttons refresh.

Kind regards
Ian

PR #147, opened 2022-02-14
Fix: 'Ghost' songs in song list

iv-gha:
Hello Gareth,

I mentioned a while back that I had 'ghosts' songs in my song list when testing.  I think I have found the problem - SongIds.txt being partially overwritten resulting in 'ghost' songs at the end of the file.  This commit always creates a new file.

I also send the user to the Welcome song after a change of storage location - seems sensible.

Hopefully get's to you for inclusion in the next release - looks like you are about to do one.

Kind regards
Ian


thebigg73 (approved):
Agree with this - I actually had a new songIdFile getting created each time in MaterialApp - can't remember if this was accidental or not!
I was ready to release the update on Saturday, but got timed out and decided to wait until after Sunday in case people were needing it for Sunday services, etc. 
I'll build shortly and update.
Thanks again Ian!

PR #148, opened 2022-02-23
Fixes: Including file write improvements

iv-gha:
Hello Gareth,

We have recently dealt with SongIds.txt being partially overwritten... I remember Alistair and others reporting corrupt song files... 

We perhaps now know that our outputstream file writes can do a partial overwrite if a longer file is re-used.  Please consider a fix for  StorageAccess.lollipopCreateFileForOutputStream that makes it create new files (no re-use) - this might globally remove the risk of OpenSongApp being a cause of partially overwritten/corrupt files?  Although I do not understand how/if this is happening!  A new file certainly cannot be partially overwritten and so it seems a good choice.

Some other enhancements and fixes too!  Continuous improvement eh!

Kind regards
Ian

P.S. I have seen truncation when exporting songs as text files - as this is affected by mode and chord display a file can become shorter on a subsequent export - and so took to looking at possible causes


thebigg73 (approved):
Hi Ian,
Thanks for your commits.  I've pulled them in and will test them out.  I've actually changed the logic of lollipopcreatefile for MaterialApp - I now pass in a boolean to delete or leave existing files.  I'm getting really close - a bit to fix on second screen display then Nearby.  After that it should be ready for release to alpha!

iv-gha:
WARNING - I think I have introduced a problem - i think it is clearing the set folder (no test if already exists I suspect) looking now!  Apologies.

PR #151, opened 2022-03-23
Enh: String changes missing from recent commit

iv-gha:
Hello Gareth,

I missed some string changes.  This asks the user to select an 'OpenSong' folder and to create a new one if needed.   This seeks to position the app as the opener of 'OpenSong' folders - the user may have more than one, they can choose to sync, copy from another device etc.  For all these cases the app continues to open a selected 'OpenSong' folder in shared storage which is never part of the app.

On another point - I noticed a small typo in the store description.  

A song book app designed for **and** musicians, singers, worship leaders, 

Is it pragmatic to add songbook to the app title?  'OpenSongApp - Songbook', as a strap line of a core feature
A user searching for a songbook will not see the app as it is way down the list - this could help.

Kind regards
Ian

p.s. you will see some NBSP in the changes - My Android Studio has decided it will not render strings with double spaces in the preview!  Is yours the same?

### DIFF
--- a/app/src/main/java/com/garethevans/church/opensongtablet/Transpose.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/Transpose.java
@@ -103,14 +103,14 @@ class Transpose {
     private final String[] fromchordsnum  = "├2┤ ├5┤ ├7┤ ├W┤ ├Y┤ ├1┤ ├3┤ ├4┤ ├6┤ ├8┤ ├9┤ ├X┤".split(" ");
     private final String[] tosharpchords1 = "»A# »C# »D# »F# »G# »»A »»B »»C »»D »»E »»F »»G".split(" ");
     private final String[] toflatchords1 =  "»Bb »Db »Eb »Gb »Ab »»A »»B »»C »»D »»E »»F »»G".split(" ");
-    private final String[] tosharpchords2 = "»»B »C# »D# »F# »G# »»A »»H »»C »»D »»E »»F »»G".split(" ");
+    private final String[] tosharpchords2 = "»A# »C# »D# »F# »G# »»A »»H »»C »»D »»E »»F »»G".split(" ");
     private final String[] toflatchords2 =  "»»B »Db »Eb »Gb »Ab »»A »»H »»C »»D »»E »»F »»G".split(" ");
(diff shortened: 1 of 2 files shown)


iv-gha:
Hello Gareth,

For Transpose a reversion of a Dec 24 change to German format behaviour.  Taking a further look at the older code it does seem to support use of A# and Ais/ais before the re-work.  This should make it transpose like the older code.  We hope!

Kind regards
Ian

thebigg73:
Hi Ian,

Thanks for your changes
_NBSP in the changes_ 
I didn't see these in the string via GitHub - don't know if it strips them out?   I actually hadn't noticed the double space issue until you pointed it out, but same with me.  A simple way around this is to quote the string
`<string name="test">"A string with          lots of spaces"</string>`

Good call with the store listing - I actually think you suggested this before (or something like it) and I forgot!.

Thanks for your fix for the transpose - I really need to spend more time getting my head around your much improved code!

Do you feel this (along with the previous pull request) are worthy of a new update on the Play store?

Gareth

thebigg73 (approved):
Pulled in commit

thebigg73:
> Hi Ian,
> 
> Thanks for your changes _NBSP in the changes_ I didn't see these in the string via GitHub - don't know if it strips them out? I actually hadn't noticed the double space issue until you pointed it out, but same with me. A simple way around this is to quote the string `<string name="test">"A string with lots of spaces"</string>`
> 
> Good call with the store listing - I actually think you suggested this before (or something like it) and I forgot!.
> 
> Thanks for your fix for the transpose - I really need to spend more time getting my head around your much improved code!
> 
> Do you feel this (along with the previous pull request) are worthy of a new update on the Play store?
> 
> Gareth

Hahaha - GitHub stripped out the multiple spaces in my example code.  The quotes work though!

iv-gha:
Hi Gareth,

I think happily there are a few fixes in this for reported faults so it
would be good to release.

However,  I"m pretty sure that capo transpose for German A B H is not good
- I suspect code is auto-identify as English for the capo line even when
set to songs are always in the chosen (German) format.  I can take a look
tomorrow - see what I can see.  Darn it transpose is tricky!

Please have a good check of the other changes - hopefully no unintended
problems!

Thanks for considering.

NBSP can be seen in the string editor.  Thanks for the tip, I can quote the
strings that use double space.  They will probably fix the display problem
but no harm in quoting anyway. Eh

Kind regards
Ian

PR #152, opened 2022-03-26
Simplification of chord detect/preferred logic

iv-gha:
Hello Gareth,

This commit changes to chord detect / preferred format use logic in the song load only, there is no need to do it again.  This removes detect calls from lots of other places.  As the chord/preferred settings cannot change doing it once is good. 

A user wishing to run without detection (they know they only want one format) can set a preferred format and let transpose use the same preferred format.  We know, because the logic is only in song load, preferred is used and detect is not run.

For a recent issue with German transpose - setting preferred to German and transpose to preferred would have 'in' and 'out' as German which will reliably work.  German format is too close to Standard to rely on detect.

It is less code!  Please consider.

Kind regards
Ian

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -466,7 +466,7 @@
 
         <activity
             android:name="com.garethevans.church.opensongtablet.PresenterMode"
-            android:configChanges="keyboard|keyboardHidden|orientation|screenSize"
+            android:configChanges="orientation|keyboard|keyboardHidden|screenSize|navigation"
             android:launchMode="singleTop"
             android:windowSoftInputMode="stateUnchanged" >
 
@@ -475,7 +475,7 @@
 
         <activity
             android:name="com.garethevans.church.opensongtablet.StageMode"
-            android:configChanges="orientation|keyboard|keyboardHidden|screenSize"
+            android:configChanges="orientation|keyboard|keyboardHidden|screenSize|navigation"
             android:launchMode="singleTop">
 
 
--- a/app/src/main/java/com/garethevans/church/opensongtablet/OptionMenuListeners.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/OptionMenuListeners.java
@@ -1123,8 +1123,7 @@ public class OptionMenuListeners extends AppCompatActivity implements MenuInterf
                 ShowToast.showToast(c);
             } else {
                 try {
-                    // Detect and use the existing format for a 0 transpose
-                    Transpose.checkChordFormat();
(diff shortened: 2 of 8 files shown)


thebigg73 (approved):
Looks sensible Ian.
Good spot on the 'navigation' flag in the manifest - I'd never heard of that one!

iv-gha:
Thanks Gareth.

I think we have

The app now prompts to move an OpenSong folder to a different storage location if the current location is not suitable.

In set lists the 'tick' button now also selects the first song. You can still select any song, including the first, from the song list.

When editing a Highlight the 'Cancel' button will discard all changes - previously a delete action could not be cancelled.

A received song is indicated with a downward pointing triangle 'download' character before the song name in the menu bar.
A song variation or slide that is stored in the set is indicated with a three bar 'set' character before the song name in the menu bar.

Set list export to text now uses a comma separated format which can be imported into a spreadsheet.

Improvements to German chord format transpose.

And lots of, sometimes sizeable, bug fixes.

Quite an eclectic list to check!

Kind regards
Ian

thebigg73:
Regarding the CSV export, I'm wondering if for MaterialApp I should include this as a separate option?  The text format is also used to build the export message (if an email program is used for sharing).  Blank fields end up showing as multiple commas in the message, whilst good for CSV fields, is bad for text readability as shown below

A love so dear, Gareth Evans, #378, , A
A safe stronghold our God is still, Martin Luther, , , C
A wonderful Saviour (He hideth my soul), Fanny J. Crosby, , , D
Abba Father, Dave Bilbrough, #133, , G

I think I'm right in saying the txt creation skipped these.  

I wouldn't worry too much about changing this for the master app - I am dangerously close to releasing an alpha test version.  Whilst not fully functional, it's almost there!  Only standard OpenSong files on both devices are currently working with Nearby, Presenter mode and Casting isn't finished yet either - I had just about finished this then had a rethink which broke it!  I'd appreciate you (and other select individuals) testing this soon - I've be doing it by sending emails to invite individuals, although I know you can test via GitHub easy enough, most won't.

iv-gha:
Hello Gareth,

No problem.  The text as CSV is to allow a person responsible for CCLI reporting to collect set list exports sent to them by others.  Loading these into a spreadsheet allows totals etc. which can simplify periodic reporting to CCLI.  A replace(", , ",", ") could reprocess it for the body text of an email and we might attach the CSV format.

I read your example as having some songs that need CCLI numbers! :-)

Alpha sounds good.  I can continue to give a focus on usability for performance/rehearsal - using the app at arms length and under pressure demands careful UI design - core functions ideally only one step away!

Kind regards
Ian

PR #154, opened 2022-04-25
Fixes and enhancements

iv-gha:
Hello Gareth,

I am suggesting moving search button and adding a set button at the top of song menu.  This pre-stages 'Material' by getting users used to accessing these features from the song menu.  Removing search from the main menu also stops the overflow menu kicking in on small screens. - this was collapsing the options button.   A user now has the options button to open the options menu (rather than having to access it from the overflow menu) - great.

Others:
On set load the 'tick' button action will open the first song if the current song is **not** in the set.
Observed a problem with song edit and ChordPro leading spaces - so fixed it.
Song menu 'flicker' observed (double call of prepare). Fixed by consolidating logic into and improving PrepareSongMenu code.

Please consider.

Kind regards
Ian


thebigg73 (approved):
Thanks Ian.  I'll pull in and test (and also look at a couple of other reported bugs).

PR #156, opened 2022-06-09
Fix: Bugs of not requesting rebuild of song menu

iv-gha:
Hello Gareth,

Found more cases where re-use of the song menu happens when it should not - I have reversed the default of prepareSongMenu to be full rebuild and changed code so that we explicitly request re-use.  This is safe as a forgotten request to re-use means a full rebuild which is still good.  Previously a forgotten request for full rebuild resulted in re-use which is bad.  This way round we are more bug tolerant!   A few other fixes and bits too!

Kind regards
Ian

### DIFF
--- a/app/src/main/java/com/garethevans/church/opensongtablet/BootUpCheck.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/BootUpCheck.java
@@ -15,6 +15,7 @@ import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
 import android.net.Uri;
 import android.os.AsyncTask;
+import android.os.Build;
 import android.os.Bundle;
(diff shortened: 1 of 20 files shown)


iv-gha:
Conflict is about

if (foundsongtitle == null || foundsongtitle.equals("")) {

I test for "" as it is being returned for some reason and would be left unchanged without this test.

Cheers
Ian

iv-gha:
Hi,

Can you consider what is needed to cause the song menu to update with a 'newly arrived from nearby' non opensong file?  This is perhaps best at the end of loadSong.  At the moment they do not appear in the song menu. and I am not sure of how to fix.

38953b1 is a fix for opensong files to appear.

Kind regards
Ian

thebigg73 (approved):

PR #157, opened 2022-06-24
Fix: LoadSong crash, Bold text fail. Enh: BootUpCheck, Options menu.

iv-gha:
Hello Gareth,

Crash and bolding fixes.  Also a menu style improvement(?).  I was also looking into not using uppercase in OPTIONS items and this is coded (not active) in OptionMenuListeners.java setxxxx voids

Example 

Current code for uppercase

    private static void setTextButtons(Button b, String text) {
        b.setTextSize(textSize);
        b.setText(text.toUpperCase(StaticVariables.locale));
        //b.setText(text);
        //b.setAllCaps(false);
    }

Make code revision for lowercase

    private static void setTextButtons(Button b, String text) {
        b.setTextSize(textSize);
        //b.setText(text.toUpperCase(StaticVariables.locale));
        b.setText(text);
        b.setAllCaps(false);
    }

Do we go to lowercase items in the options menu?  For me less of a SHOUT.

Kind regards
Ian


thebigg73:
Happy to go lowercase in menus, although this is obsolete in MaterialApp anyway. 
The only potential issue is that some of the strings for the menus (I think) are stored in CAPS in the strings.xml file, so we may end up with a mix without manually changing them
NewMenus

thebigg73:
I've removed allcaps from the menu items, but added the textsize back in as it was a little small otherwise

iv-gha:
Hi, not sure why that happens, I scale lines by 0.8 and compensate by
taking the user preference size and multiplying by 1.25 and  used.  The
result should be the same effective size as before. Perhaps 1.25 needs to
be slightly larger to stop rounding down.

Odd.

Cheers
Ian

iv-gha:
Hi,

In OpenMenuListener,java...

    static void optionListeners(View v, Context c, Preferences preferences, StorageAccess storageAccess) {

        // IV - Option menu lines are scaled by 0.8, multiple by 1.25f to scale text font
        textSize = 1.25f * preferences.getMyPreferenceFloat(c,"songMenuAlphaIndexSize",14.0f);

The 1.25f could be made larger to support bigger menu text compared to song menu text.  This is part of the logic that eventually overrides text size to match the user's preference (all those setxxxx calls, as mentioned previously).  The layouts do not (always) look like the final thing - as preference sizing is used.

Kind regards
Ian

Edit: 1.40f looks pretty  good to me, when 17sp is the preference - this only affects the Options menu display.  Song menu uses the preference size as-is.
What was the case for scaling this Options menu?  If we want fixed then setting textSize above to the wanted size would do it.  I guess it is good for accessibility to retain it.

thebigg73:
Hi Ian,

Apologies - my stupidity! I thought that you'd commented out the setTextSize likes on the option menu text where you'd suggested disabling the all caps.  That's what I thought I'd reverted - it turns out you hadn't and the sizing is fine!!

PR #177, opened 2022-08-21
Set list A-Z and highlight of song

iv-gha:
Hi Gareth,

I tweaked  the  A-Z sort to just sort on song name and not folder/song name. This stops grouping of songs by folder - otherwise variations etc. are out of sequence!

I also added a highlighting of the current song icon in a set to make it  more obvious where you are.

A few post testing changes and I think it is there.

Have a look.

Cheers
Ian

### DIFF
--- a/app/src/main/java/com/garethevans/church/opensongtablet/PopUpSetViewNew.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/PopUpSetViewNew.java
@@ -447,6 +447,7 @@ public class PopUpSetViewNew extends DialogFragment {
             helpSwipeItem_TextView.setVisibility(View.GONE);
             listSetTweetButton.hide();
             set_shuffle.hide();
+            sort.hide();
             helptext.setVisibility(View.VISIBLE);
         }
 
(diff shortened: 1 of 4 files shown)


thebigg73 (approved):
I did consider that sort, but I couldn't figure out the comparator bit and was going to have to adjust the set string to filename/folder, sort, then adjust it back again.  Well done for figuring that out!
I really like the highlighting of the set current item.  However, I know from previous work that the highlight colour set in the recyclerview this way will mess up when scrolling up and down.  I've fixed in MaterialApp and will adapt the code here.  Basically you have to use a separate sparseBooleanArray that tracks the highlighted option as the position doesn't always returm the correct value.

iv-gha:
This 'take' on it is done in onBindViewHolder with the icon dynamically set to low or high - might mean the scroll problem is not there!

EDIT: Ahh! I see it breaks if the songs are re-ordered abd you scroll!  Test instead agasint si.songitem seems to fix this in 'master'...

        // IV - Highlight icon of current song when in the set
        if (si.songitem.equals((StaticVariables.indexSongInSet + 1) + ".")) {

PR #178, opened 2022-08-22
SetListAdapter - Highlight/Lowlight v3

iv-gha:
Hello Gareth,

I have combined bits of my proposed change and your changes to fix stability of song highlight under conditions of set re-order/song deletes/scrolling - stuff users do!  It depends on the displayed number of the song remaining unchanged as a highlighted song is moved.  I believe we can get away without an independent list on this occasion.  Hopefully caught all edge cases!

Cheers
Ian


thebigg73 (approved):
Pulling in to test

PR #179, opened 2022-09-28
Fix: Crash fix - 'makeVariation'

iv-gha:
Found a crash!

### DIFF
--- a/app/src/main/java/com/garethevans/church/opensongtablet/PopUpSetViewNew.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/PopUpSetViewNew.java
@@ -34,7 +34,6 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
-@SuppressWarnings("ComparatorCombinators")
 public class PopUpSetViewNew extends DialogFragment {
 
     private static Dialog setfrag;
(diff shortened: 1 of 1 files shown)


iv-gha:
Hi Gareth,

Ability to add song quickly to a set as a variation using long press of '+'...

So on a Sunday meet, I have a set to share with projectionist and other musicians / singers.  We all connect.  They create a new set and connect to receive my songs (no keep).   Once connected, I step through my songs, they long press '+' to add them to their set.  At the end they all have a set in the correct order of variations with these being my songs (my key, my ordering, etc) with no effect on their own song library.  Done with long press '+' actions.

If during performance connection issues occur then everyone has my songs in a set to use.  Safe!

I could have prepared and shared a set of variations to get this, but that is painful and for many leaders a seemingly big ask!  The above is dynamic allowing any OSA user to collect the leaders set (with any last minute changes, different keys, different order etc) on the fly.  It has no need for sync of songs, use of email etc - Just OSA.

I think Material would not as easily support this yet.

Hence the fix to keep my Master friends happy and able to do this!

thebigg73:
As if I'd forget about your amazing contributions.....

I've added a page button.  Short press=add to set, long press=add to set as
a variation.

[image: addtosetbutton.png]
[image: addtosetvariation2.png][image: addtosetvariation3.png]

iv-gha:
Sounds good.  This fix needs test and release as '+' long press is broken in 'Master'.

Cheers

thebigg73 (approved):
Thanks!

PR #181, opened 2022-11-08
Fixes

iv-gha:
HI Gareth,

Fixes.  Please see commit comments for details.

Kind regards
Ian


app/src/main/java/com/garethevans/church/opensongtablet/MenuHandlers.java
```diff
@@ -76,10 +75,8 @@ static void forceOverFlow(Context c, ActionBar ab, Menu menu) {
             ViewConfiguration config = ViewConfiguration.get(c);
             if (Build.VERSION.SDK_INT<Build.VERSION_CODES.R) {
                 Field menuKeyField = ViewConfiguration.class.getDeclaredField("sHasPermanentMenuKey");
-                if (menuKeyField != null) {
-                    menuKeyField.setAccessible(true);
-                    menuKeyField.setBoolean(config, false);
-                }
+                menuKeyField.setAccessible(true);
+                menuKeyField.setBoolean(config, false);
```
thebigg73: This is the only bit of the pull I'm a bit wary of.  This is a 'hacky' way to check for hardware back buttons on phones and from memory, it quite regularly throws a null menuKeyField value as not all versions of Android have this declaredField.  Every example online has the null check before setting values. 
Given that most devices don't have hardware devices, it is probably more sensible to either leave the null check, or remove this check altogether (as I have in v6).

thebigg73 (changes requested):
Hi Ian,

Thanks so much for your excellent contributions.  I've flagged one issue that I think needs changed (the menu hard key in the MenuHandlers file (I think you should be able to see my comment on the file?). If not, here's what I put:

This is the only bit of the pull I'm a bit wary of.  This is a 'hacky' way to check for hardware back buttons on phones and from memory, it quite regularly throws a null menuKeyField value as not all versions of Android have this declaredField.  Every example online has the null check before setting values. 
Given that most devices don't have hardware devices, it is probably more sensible to either leave the null check, or remove this check altogether (as I have in v6).

app/src/main/java/com/garethevans/church/opensongtablet/MenuHandlers.java
iv-gha: Code Inspector suggests that this is not needed - but what does it know! I have reverted. :-)

PR #182, opened 2022-11-11
Enh: Check device location services are ON,  before opening connect devices. Export set list comma fix

iv-gha:
Hi Gareth,

Another pain point for me...

One of my music friends uses OpenSongApp on a Fire 10 HD and seems somehow to keep turning Location Services off!  This stops their use of 'Act as Client' from ever connecting when we are sharing songs.  Fixing involves painful searching for the Location options in strange settings options!

This commit tries to test for Device Location Services being ON and if not have the user correct before using Connect Devices, opening the Location settings for the user to adjust.    Works for me on Android 12, 8, 5.  Found the approach on the Internet and used a similar dialog style to other 'Connect devices' dialogs.

Please consider.

Kind regards
Ian

### DIFF
--- a/app/src/main/java/com/garethevans/church/opensongtablet/ExportPreparer.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/ExportPreparer.java
@@ -456,7 +456,8 @@ class ExportPreparer {
                                         // Read in the song title, author, copyright, hymnnumber, key
                                         getSongData(c, songuri, storageAccess);
                                     }
-                                    sb.append(song_title);
+                                    // IV - Handle comma in name - remove comma with no following space as used in numbers - '10,000 Reasons'
+                                    sb.append(song_title.replace(", "," | ").replace(",",""));
                                     sb.append("¬ ");
                                     if (!song_author.isEmpty()) {
                                         sb.append(song_author);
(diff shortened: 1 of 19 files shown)


thebigg73 (commented):
Looks good Ian.   I'm unable to test for the permission not being granted, but happy that the alert can be shown.  Can you confirm that the location check line is correct:

`if (!gps_enabled && !network_enabled) {
        ...
}`

I'm wondering if it should be an or comparison:
`if (!gps_enabled || !network_enabled) {
        ...
}`

iv-gha:
Hi,  I looked at that...  the code is testing for location is off and if so asks for it to be put on.  I understand the logic as, if gps is off and network is off then no location features are in use and so location is determined as off.  I think that is what is going on - I too thought it looked odd when coded.

thebigg73:
Thanks for checking!  I've merged these changes to MaterialApp too (which only needs custom image slides finished off in Presenter Mode then it should be ready for promotion to beta...)
I'll pull this in and get an update to the Play Store tonight.

Thanks again Ian.

iv-gha:
On further thoughts...  needs adjustment...

For Location, Nearby needs network enabled and does not make use of GPS.  I think on an Android 5 device which has GPS (my tablet does not) then Location ON and in Device only mode will have GPS only enabled.  The current test would say this is OK but Nearby client mode would not work as Location does not have network enabled.

So, I think our need is only to test for Location having network enabled.  I have added a commit.  Please take a look at the rework.

PR #183, opened 2022-11-14
Enh: Revised check of location services before opening connect device…

iv-gha:
…s to only NETWORK_PROVIDER.


thebigg73 (approved):
Looks sensible.  I've changed a bit of the permissions around nearby as well that I'll include,. 
 Having tested with API33 on the emulator, I noticed that BLUETOOTH_ADVERTISE is also required (in addition to the nearby permission).  I don't know if this might be related to your issue

PR #184, opened 2022-12-01
Nearby permission - manual adjust backstop

iv-gha:
Hello Gareth,

When testing, I spotted that when a user has, in some way, opted to not allow permissions, 'Connect devices' will not open and no prompt is made to allow permission - the user needs to manually adjusting permissions to fix this.  I have added a prompt to manually adjust permissions that takes the user to OpenSongApp 'App info' and simplified some code.  Hopefully good across Android versions!

Also, I have tried to improve Nearby 'Broadcast name' - the 'random' name is no longer seen the first time 'Connect devices' is used after a prompt for permissions.

Please consider.

Kind regards
Ian

### DIFF
--- a/app/src/main/java/com/garethevans/church/opensongtablet/NearbyConnections.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/NearbyConnections.java
@@ -34,7 +34,6 @@ import java.io.OutputStream;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Objects;
-import java.util.UUID;
 
 public class NearbyConnections implements NearbyInterface {
 
@@ -198,21 +197,10 @@ public class NearbyConnections implements NearbyInterface {
     }
 
     public String getUserNickname() {
-        try {
-            if (StaticVariables.deviceName == null || StaticVariables.deviceName.isEmpty() ||
-                    (FullscreenActivity.mBluetoothName != null && FullscreenActivity.mBluetoothName.equals(StaticVariables.deviceName))) {
-                if (FullscreenActivity.mBluetoothName == null || FullscreenActivity.mBluetoothName.equals("Unknown")) {
-                    FullscreenActivity.mBluetoothName = UUID.randomUUID().toString().substring(0, 8);
-                    FullscreenActivity.mBluetoothName = FullscreenActivity.mBluetoothName.toUpperCase(StaticVariables.locale);
-                }
-                StaticVariables.deviceName = preferences.getMyPreferenceString(context, "deviceId", "");
-                // IV - If the user has not set an override name use the Bluetooth name
-                if (StaticVariables.deviceName.isEmpty()) {
-                    StaticVariables.deviceName = FullscreenActivity.mBluetoothName;
-                }
-            }
-        } catch (Exception e) {
-            e.printStackTrace();
+        StaticVariables.deviceName = preferences.getMyPreferenceString(context, "deviceId", "");
+        // IV - If the user has not set a device name use mBluetoothName
+        if (StaticVariables.deviceName.isEmpty()) {
+            StaticVariables.deviceName = FullscreenActivity.mBluetoothName;
         }
         return StaticVariables.deviceName;
     }
--- a/app/src/main/java/com/garethevans/church/opensongtablet/NearbyConnections.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/NearbyConnections.java
@@ -34,6 +34,7 @@ import java.io.OutputStream;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Objects;
+import java.util.UUID;
 
 public class NearbyConnections implements NearbyInterface {
 
@@ -197,10 +198,21 @@ public class NearbyConnections implements NearbyInterface {
     }
 
     public String getUserNickname() {
-        StaticVariables.deviceName = preferences.getMyPreferenceString(context, "deviceId", "");
-        // IV - If the user has not set a device name use mBluetoothName
-        if (StaticVariables.deviceName.isEmpty()) {
-            StaticVariables.deviceName = FullscreenActivity.mBluetoothName;
+        try {
+            // IV - If deviceName is null, determine the current deviceName.
+            if (StaticVariables.deviceName == null) {
+                // IV - If user has set a device id then use, otherwise use Bluetooth name. If no Bluetooth name (should never happen), use a random name.
+                StaticVariables.deviceName = preferences.getMyPreferenceString(context, "deviceId", "");
+                if (StaticVariables.deviceName.isEmpty()) {
+                    if (FullscreenActivity.mBluetoothName != null) {
+                        StaticVariables.deviceName = FullscreenActivity.mBluetoothName;
+                    } else {
+                        StaticVariables.deviceName = UUID.randomUUID().toString().substring(0, 8).toUpperCase(StaticVariables.locale);
+                    }
+                }
+            }
+        } catch (Exception e) {
+            e.printStackTrace();
         }
         return StaticVariables.deviceName;
     }
--- a/app/src/main/java/com/garethevans/church/opensongtablet/OptionMenuListeners.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/OptionMenuListeners.java
@@ -28,6 +28,8 @@ import com.google.android.gms.common.ConnectionResult;
 import com.google.android.gms.common.GoogleApiAvailability;
 import com.google.android.material.floatingactionbutton.FloatingActionButton;
 
+import java.util.Objects;
+
 public class OptionMenuListeners extends AppCompatActivity implements MenuInterface{
 
     @SuppressLint("StaticFieldLeak")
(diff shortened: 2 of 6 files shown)


thebigg73 (approved):
Thanks for these Ian.  I've not ported them to Alpha as they are dealt with in a slightly different way already that seems to work.

iv-gha:
Sounds good.

You may like to consider 'backstop' behaviour in Alpha.  Outside of OpenSongApp, deny permissions and see how Alpha then behaves.  Hopefully it always offers the user a way forward.

On another note, I see the work on italics and brackets - my thought is that text in round brackets is not displayed when projecting as they are comment.  I use square brackets for projecting things like call and [answer] lines.  In effect you give an option to make in-line comments on lyric lines italic.

Cheers

thebigg73:
Hi Ian,

Alpha uses a bottom sheet to deal with the backstop issue.  I've tested a
few scenarios and it seems to fire up.  The settings take you to the app
settings page:
[image: backstop.png]

Italics/brackets - they should be displayed when presenting?  Only comment
lines (starting with ;) are ignored.  I wanted to avoid square brackets as
I suspect it'll cause big issues if the user chooses to edit the song in
ChordPro mode.  I did think about curly brackets { }, but figured that many
users will have already used normal brackets to add notes, echos,
harmonies, etc. and this would give the option of formatting them without
having to edit their song.
[image: performancebrackets.png]
[image: presenterbrackets.png]

Are you meaning you put answer lines as comment lines?
 The normal line
;[The answer line]

thebigg73:
I should've also said that I edit the song to display echos on new lines in
presenter mode using the | character:[image: editsongbrackets.png]

iv-gha:
I am thinking of second screen stuff...

When projecting lyrics (no chords) on a second screen then round bracket text, comment, is removed and white space is replaced with a single space and ' - ' removed. The line is trimmed.

So

```
.    A    G    A
     Our  fa - ther
.    A    G    A
     (Our Fa - ther)
```

Would project as

Our Father

Note no second line.

and

```
.    A    G    A
     Our  Fa - ther
.    A    G    A
     [Our Fa - ther]
```

projects  as

Our Father
[Our Father]

As you say, '|' could be involved.

```
 Our  Fa - ther| [Our Fa - ther]
```
If a user wants to highlight using brackets then round brackets cannot be used when wanting to project the brackets and text if projecting lyrics only.

Hopefully I am clearer! :-)

thebigg73:
Hi Ian,

Is this something we added to v5?  v6 displays absolutely fine:
[image: brackets1.png]
[image: brackets2.png]

There is nothing in OpenSong format that says a comment line starts with a
bracket.  Comment lines should only start with ;

thebigg73:
Hi Ian,
I must be missing something - booted up v5 and that also seems to display
fine using your sample.

[image: brackets3.png]

iv-gha:
HI Gareth,

Yes, you are right there was a v5 change 

https://github.com/thebigg73/OpenSongTablet/commit/42befc50b75c0c262bfd54a6833bf6b107e99e2f

on  26th Feb 2021

This addressed use of '(....)' in lyric lines for optional words and instructions.  For example

(Well) He is the Lord (to Chorus 2)

It also addressed hyphen used to split words for a mid word chord change.  Only hyphens with spaces around are removed.

The heav -   en   - ly host

These forms are common in the high quality song layouts provided by SongSelect.

Where it causes a problem is for use of brackets in call and answer songs, the song needs to be edited to use square brackets if projected.  Our projecting friends have been used to this behaviour since 2021.

Here is one example of the logic.  It is in three places, search for

Remove (....) comments

```
                    // IV - Remove (....) comments when Stage or Presentation mode song
                    if (!StaticVariables.whichMode.equals("Performance")) {
                        if (!StaticVariables.whichSongFolder.startsWith("../") || StaticVariables.whichSongFolder.startsWith("../Variation")) {
                            bit = bit.replaceAll("\\(.*?\\)", "");
                        }
                    }

                    // IV - Remove typical word splits, white space and trim - beautify!
                    bit = bit.replaceAll("_", "")
                            .replaceAll("\\s+-\\s+", "")
                            .replaceAll("\\s{2,}", " ")
                            .trim();
```
These replaces affect lyrics only display and lyrics only projecting.

Cheers
Ian

thebigg73:
Thanks for highlighting this code!  I wasn't seeing this issue as I was
failing at the if (lyricsOnly) {..} stage as the following line:
boolean lyricsOnly = ((presentation && !presoShowChords) || (!presentation
&& !displayChords));

was always returning false as I hadn't specifically switched off
presoShowChords and it was somehow defaulting to false (I don't know if
that is something worth looking at given that I think alpha is just about
ready to push to beta - I just need to finish testing the Nearby logic).
When I switched these on, then off again, it then caused lyricsOnly to
trigger as true and then hid the lines as you described.

I notice that as expected, if you use square bracket comments, this
completely breaks the edit as OpenSong/ChordPro option.  Toggling back and
forward progressively ruins the song!  What might be sensible to do is to
programmatically change [..] comments to {...} comments on save/open in
OpenSong format, meaning that curly brackets should be used going forward
as an additional option to regular brackets, but any songs loaded with
square brackets will be silently converted and processed as curly
brackets.  I definitely wouldn't want bracketed text not be be available to
show on presented screens - I've just never noticed that behaviour to this
point.

I think I might look at adjusting the bracket option to allow users to set
(..) comments as formatted comments and curly brackets as note only
comments that aren't projected.  This would allow both features to remain.
How does that sound?

iv-gha:
Hi Gareth,

We could just suppress in-line comments (but not lines that are completely bracketed)

```
(....) First lyric (....) line (....)
     (Second lyric line)
```

Suppressing in-line (....) only would give

First lyric line
(Second lyric line)

This would mean

```
 (....) A lyric (....) line (....)
 (To END last time)
```

Would need edit to change the instruction line...

`;(To END last time)`

if used for project of lyrics only.

An option could be added to control the replace of 'in-line (....)'.  It would need to default to on.  If we go this route then we should discuss other options for hyphen replacement etc  too.

This would always support project of '(...)', with or without the italics option, where it is a complete line.

I think opensongapp specific use of characters, like you propose for '{', moves away from the Opensong  file format.  I do not think we should do this.

What is proposed above is OpenSongApp handling how it displays a song that is in OpenSong format. This approach is similar to ChordPro which sets the format of a song file but does not dictate how the app interprets this for display / print.  An app can use or ignore directives as it pleases.  Clearly ChordPro, being aimed at printing, has more formatting directives - whilst OpenSong has less being aimed at projection.

Sorry for the ramble.

Please discuss.

Kind regards
Ian

thebigg73:
I'm pretty sure that bracketed lines are not a feature in OpenSong format
for comments, unless I've missed something?  The only commenting option is
to use a line with a semicolon.

I regularly have comments in brackets that I do want to display in
presentation mode as instructions for the congregation and have added them
into songs with the assumption they weren't hidden (I've been using alpha
for over a year so didn't really appreciate this change to the v5 code and
or projectionist uses desktop OpenSong).  I think if people want text to
disappear in projection mode when projecting in deep OpenSong or
OpenSongApp they should use comments for full lines, or some non- standard
formatting such as {..} for inline text, but they would have to accept that
the desktop song would ignore this and display the text regardless.

Either way with round brackets to hide text or curly brackets to hide text,
neither are OpenSong formatting and users are less likely to have
accidentally put text in curly brackets mid lines, so shouldn't have
disappearing text they weren't expecting.  We normally project with the
desktop app so all bracketed text shows up there and only proper ;comment
lines are not displayed.


Alpha already has logic for hyphens and other rogue text (word trimming).

iv-gha:
Yes, I understand.

The aim of the earlier commit, maybe oddly, was to avoid changing lyric content when using SongSelect.  That is to support legal use of SongSelect songs with good display for musicians, singers and for projection with minimal edit of lyrics when using OpenSongApp alone.

OpenSongApp imports from SongSelect. The import initially gives certainly legal lyrics and copyright info.  A song edit can then be used to edit tempo, capo etc and add | and || for projection.  No change to lyrics is permitted, change of line splits is up to the user.

It is the nature of optional words and instructions that they are not mandatory, or even intended, for projection.  
The user would need to edit SongSelect imports to prevent optional words and musical instructions being projected.  They must be careful not to change lyrics as this makes use illegal.

The behaviour that suppressed (....) and the use of ' - ' on lyric lines means a SongSelect song can be imported and put in-use with minimal editing and high certainty of legality.

Obtaining copyrighted song lyrics from non official sources requires confirmation of all lyrics, copyright details and right to legally use by check of SongSelect (or other service providing legal use of lyrics) - obviously it is easier to get songs from SongSelect.

Having said all that....

**** I will see how things are if the '(....)' replace is removed.  It might have been a change too far! ****

Cheers
Ian

thebigg73:
In that case, would it be worth looking at changing the import SongSelect
feature (I'd suggest for alpha) to replace the inline comments that
shouldn't be projected that you mention to format them in curly brackets.
Given importing from SongSelect is only a feature of OpenSongApp (not
OpenSong), the curly brackets can then be set to ignore in presentation
mode and in the desktop app, they will just appear in the brackets.  Users
can then choose if they wish to manually remove them.

I currently don't have access to a SongSelect account so can't test this
just now other than using a small number of freely available songs on their
site (public domain).  Would this be a sensible compromise?

PR #187, opened 2023-02-15
fix folder import

millerthegorilla:
Using an old backup, I tried to import my songs, and was faced with spurious directory creation, at various points in the directory tree (under Songs), that were empty.   So, I hacked the code and added some existence tests, and the directory tree structure and files are all created correctly, as far as I can tell, with the exception of two top level copies of the final top level directory.
This doesn't seem to occur with a newly created folder system, backed up and reimported, but I haven't created a particularly large tree to test.
I haven't got the time currently, to devote to refactoring the code, but there are various optimizations that could be made.

### DIFF
--- a/app/src/main/java/com/garethevans/church/opensongtablet/PopUpImportExportOSBFragment.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/PopUpImportExportOSBFragment.java
@@ -394,7 +394,7 @@ public class PopUpImportExportOSBFragment extends DialogFragment {
                     for (String aFoldersselectedtoimport : foldersselectedtoimport) {
                         // Is it in the main folder
                         if ((aFoldersselectedtoimport.equals(getString(R.string.mainfoldername) + "/") ||
-                                (aFoldersselectedtoimport.equals("MAIN" + "/"))) && !ze.getName().contains("/")) {
+                                (aFoldersselectedtoimport.equals("MAIN/")))) {
                             oktoimportthisone = true;
                             // Or another folder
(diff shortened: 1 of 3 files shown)


thebigg73:
Hi James,

The master branch is nearing end of life and will soon be replaced by the MaterialApp branch which is currently in beta testing.  This has been rewritten from the ground up to improve the UI and make the code more reusable and easier to maintain.  

I've rewritten the import OSB backup code completely, so I'm not sure about bringing those changes into the new version.   If you still have that old backup file, please email it to me and I'll test.
The storage access changes seem sensible, so I've incorporated them into the MaterialApp branch and will appear in beta 5

millerthegorilla:
OpenSongBackup_2021_04_19.osb
<https://drive.google.com/file/d/1aonK5Vgnwkss_EGHfebxF58ZxWQMWUG2/view?usp=drivesdk>

Hi Gareth, and thanks for getting back to me.  I am glad to hear the app is
continuing to be developed, it is quite a marvellous bit of kit.

I would be grateful if you could take a look and see if the attached
Opensong backup imports ok.

I imported it, cleaned up the structure and exported it, but when I
imported it, the same issue was present.

I had a quick look at the material app branch, but I was unable to get it
running, due to a gradle incompatibility I think.   I am mostly programming
in Python currently and will be for the next year at least, but would like
to get back into doing some Java/Kotlin at some point.
It looks to me like the new version of Opensong will be daemonised, which
is interesting.

I was considering setting aside some time to refactor the internals of the
storage access API, to streamline it and make it a little safer, re threads
etc, but i am sure it will work just fine as it is.

As always, thanks for all your hard work!

All the best

James

PR #188, opened 2023-02-23
Material song display fixes

iv-gha:
Please see commit comments.  I only expected the recent commit on 23rd to go up. My rebase on 20th seems to have not quite made me identical to you at that point - so other commits are here sorry.

These are improvements for v6 song render - I believe.  Take a look!

Happy to discuss!

Kind regards
Ian

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -69,7 +69,7 @@ android {
 dependencies {
     implementation 'androidx.activity:activity:1.6.1'                                   // Required for permission requests
     implementation 'androidx.fragment:fragment:1.5.5'                                   // Using fragments and navigation
-    implementation 'com.google.android.material:material:1.9.0-alpha01'                 // Required for material design goodness
+    implementation 'com.google.android.material:material:1.9.0-alpha02'                 // Required for material design goodness
     //implementation "androidx.compose.material:material:1.3.1"                         // Can upgrade theme to Material3 once I remove support for KitKat
     implementation 'com.google.android.gms:play-services-cast:21.2.0'                   // For casting obviously!
     implementation 'com.google.android.gms:play-services-cast-framework:21.2.0'         // For casting obviously!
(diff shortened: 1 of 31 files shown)


thebigg73 (approved):
I'm assuming the removal of matchPresentationOrder from several locations is due to adding it into the main process song logic?

iv-gha:
You are right - matchPresentationOrder is called in the process song logic as in v5.

Adding splits will 'reveal' any areas where use of presentation order is needed.  I have come across a few and am looking further.  I note presentation mode is a bit disturbed - lookig at that too.

The aim is robust song section info confirmed as available and correctly used by all modes.

Please consider the issue I have raised about missing sections on display in stage mode - this may have been revealed by processing higher numbers of sections for presentation order.  It appears the sections do not size correctly!

Kind regards
Ian

thebigg73:
Something not quite right with the song sections.  An chord line after a section (blank line in between) then gets merged into a new  section header based on the previous one.  Also happens with comment lines
Screenshot from user attached illustrating issues.

I'll see if I can fix it.  Not sure if this is due to recent changes or not though

Screenshot_20230303_144747_OpenSongApp

thebigg73:
Note the heading starting with C.  as the previous section is the chorus

thebigg73:
Found the issue in ProcessSong and pushed update - may warrant further discussion?
`// GE reverted change
        for (int x = 0; x < songSections.size(); x++) {
            fixedlyrics.append("\n§");
            if (songSections.get(x).startsWith("[")) {
                sectionHeader = songSections.get(x).substring(0,songSections.get(x).indexOf("]") + 1);
                fixedlyrics.append(songSections.get(x));
            } else {
                // GE NOT SURE THAT THIS IS USEFUL.  A separate section with or without a section
                // shouldn't take the previous section label.  If it did, it would need the "\n" first
                //fixedlyrics.append(sectionHeader).append("\n").append(songSections.get(x));
                fixedlyrics.append(songSections.get(x));
            }
        }`

PR #193, opened 2023-03-03
Material app - Fixes

iv-gha:
Hello Gareth,

A fix for double new line splits causing layout problems - refinement to ensure valid section headers when splitting.

Kind regards
Ian


thebigg73:
Hi Ian,
We were both working on this at different times!  Can you have a look at my code in ProcessSong (just pushed a few mins ago before I saw your pull request).  If the header is pushed it should have a new line after this before appending the content.

thebigg73:
Actually I see you've spotted the new line \n missing and added it.  Because there are conflicts, I'll manually merge and push this up to GitHub.

iv-gha:
Hello Gareth,

I think this needs to be with me.  There has been a lot of v5 layout logic that did not make it into v6 - as it was forked around 5.1.4 this is understandable, the more v6 diverged the harder this would have been.  Are you OK to let me lead this fix?

PR #194, opened 2023-03-03
Material app - getChordPositions fix

iv-gha:
Hello Gareth,

This fixes

```
.C
 Some lyrics
```

Note a single chord at the beginning for the line.

Laying out as
```
C
 Some lyrics
```

Note the space at the beginning of lyrics.

A quirk of v5 is that chords lines sent to getChordPositions had their first char swapped for a space.  v6 does it right so the line arrives with its line type char.  This commit fixes getChordPositions to treat the first char as space all the time - better and fixes the above glitch.


iv-gha:
I have a commit that is stuck in my branch that is the actual fix for this problem - please take a look at 
https://github.com/iv-gha/OpenSongTablet/commit/f2fac333974cb87e5f07c3202470a5a002baaa3a
I suggest you browse the commit file and copy/paste it somehow into your Android Studio to see the changes that fix ths.

The above 'x == 1' 'fix' is not the root cause of the problem and should not be applied.  Whilst it fixes the lyric line It will disturb a chord of more than one char a the beginning of the chord line.  Doh!

Apologies
Ian

PR #195, opened 2023-03-03
Enh - v5 Pad enhancements applied

iv-gha:
Had a problem with Android studio - so Pad code not quite the latest - will fix.

Hello Gareth,

v6 Pads were not stable - lots of instability as code forked when v5 was unstable.  I have applied latest v5 techniques using the v6 approach of using Timers.  Sleep is no longer needed - important as the processes run in the main thread!

The Pad time display is a single timer that runs while Pads are in use (v5 approach) to give a smooth display on pads stop/starts.

Simpler code too!

See commit comments.  Hopefully a good commit.

Kind regards
Ian

ADDED other fixes too!

### DIFF
--- a/app/src/main/java/com/garethevans/church/opensongtablet/MainActivity.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/MainActivity.java
@@ -3166,8 +3166,7 @@ public class MainActivity extends AppCompatActivity implements MainActivityInter
         nearbyConnections.turnOffNearby();
         // Stop pad timers
         if (pad != null) {
-            pad.stopTimers(1);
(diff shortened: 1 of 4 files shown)


iv-gha:
The media player does not start on cross fade every so often with error -38. when it tries to do getDuration().

It seems this is a symptom of the player not being fully prepared even though doPlay, with the start() in, is called by the players setOnPreparedListener!  v6 seems to create new mediaplayers all the time(?), did v5 reuse them?

Commit closed as I cannot figure it out!  The code correctly calls loadAndStart, but it does not then deliver a working player to doPlay()!  The effect is the cross fade does not happen.
  
Cheers
Ian.

PR #199, opened 2023-03-17
Fixes and v5+

iv-gha:
Hello Gareth,

This further refines pads and the song layout engine.

Please see commit comments.

A fix for trimSections helps song render - I am hoping I do not find more to fix with song render!  During testing I turned off trimSections to discover it still applied - doh!

Kind regards
Ian

### DIFF
--- a/app/src/main/java/com/garethevans/church/opensongtablet/performance/PerformanceFragment.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/performance/PerformanceFragment.java
@@ -158,11 +158,6 @@ public class PerformanceFragment extends Fragment {
         mainActivityInterface.getToolbar().setPerformanceMode(true);
         mainActivityInterface.showActionBar();
 
-        // Set tutorials
-        Handler h = new Handler();
-        Runnable r = () -> mainActivityInterface.showTutorial("performanceView",null);
-        h.postDelayed(r,1000);
-
         // MainActivity initialisation has firstRun set as true.
         // Check for connected displays now we have loaded preferences, etc
         if (mainActivityInterface.getFirstRun()) {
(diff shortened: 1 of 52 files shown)


iv-gha:
Hi Gareth.

You will need https://github.com/thebigg73/OpenSongTablet/pull/199/commits/ade39acedda4fee197b0505c40ba5514be7a4fbc

iv-gha:
The layout engine needed a fair bit of work - we now have working splits, various 'trim' options behaving and output usable in all modes.  I have raised an issue for capo transpose not handling flats properly.  Getting render right is key and it may need further work.  I recognise the stress of having to pick through the proposed changes.

The Pad engine and settings behaviour are I think now good.

app/src/main/java/com/garethevans/church/opensongtablet/performance/PerformanceFragment.java
```diff
@@ -800,14 +800,28 @@ public void onGlobalLayout() {
...
+                mainActivityInterface.getSong().setCurrentlyLoading(false);
+
+                // Release the processing lock
+                processingTestView = false;
+
+                // Set the positions
+                mainActivityInterface.getDisplayPrevNext().getPositions();
+
+                // Slide in
+                long QOSAdjustment = doSongLoadQOSTime - (System.currentTimeMillis() - doSongLoadStartTime);
```
thebigg73, line 818: What is 'doSongLoadQOSTime'
This doesn't appear anywhere in the pull request

app/src/main/java/com/garethevans/church/opensongtablet/performance/PerformanceFragment.java
```diff
@@ -893,12 +907,6 @@ private void dealWithExtraStuffOnceSettled() {
             displayInterface.updateDisplay("setSongInfo");
             displayInterface.updateDisplay("setSongContent");
 
-            // Send a call to nearby devices to process the song at their end
```
thebigg73, line 896: Don't think this should be deleted?

thebigg73 (changes requested):
There seems to be quite a few conflicts and strange code changes (some of which are already present in the current branch). 

As a result rather than pulling this in as it is, I've gradually manually made most of the changes proposed, with the exception of those in the ProcessSong and PerformanceFragment files - I'm not sure this pull request has fully incorporated the changes I had previously made in ProcessSong and the PerformanceFragment has missing variables and a removal of nearby song posting.
I think it might be worth creating an updated pull request

app/src/main/java/com/garethevans/church/opensongtablet/performance/PerformanceFragment.java
iv-gha: You are right.  The pull has gone odd!  It seems to have a mangled PerformanceFragment.java file.  I suggest you do not merge this file.

A challenge for me is that you are adding major changes to the beta - including new features which is undermining fixing a beta. This makes it alpha and very difficult to offer fixes for.  I suspect I did not get a merge right!

app/src/main/java/com/garethevans/church/opensongtablet/performance/PerformanceFragment.java
iv-gha: I have a commit that addresses making the app more Material.  I suspect bits have crept into this commit - sorry.
As I did not want to overwhelm you I deferred commit which is now hampering my ability to work on any other fixes.  This is a 'make it material' commit - with more material standard layout and icon use.  A similar thing happened to v5.

This is improving the beta, to get it looking good.

I can prepare a pull?

Enh: Action bar
- Fixes for variable clock size, button positions and glitches when moving around the option menu
 - Item sizes and positioning made consistent
 - Option menu transitions made cleaner.
 - Help icon moved into option menu
 - Fragment calls of updateToolBarHelp moved to onResume to ensure help button onclick action
 - Title and Battery info click is active only on top option menu
 - Battery info is 'gone' when not in use
 - Whole title area now clickable.

Enh: 'Material' styling applied.
- Flatter dialog tops
- 'settings_outline', 'help_outline' and 'web_outline' 24dp icons from google used in relevant places in app. Other navigation icons set to 24dp.  The app matched with the material styling of storage, cast and other system dialogs.

Fix: Custom margins
- Are re-applied on orientation change
- Song menu position now respects custom margins.
- An open Song menu is closed on orientation change

Enh: 30% faster song transitions
- v5 optimisations appled, song animate out is short and combines with a QOS behavior to give a user perception of consistent speed (now faster) of song change.

PR #207, opened 2023-03-24
Material app

iv-gha:
Apologies if this contains old commits you have made - Android Studio and Github are having fun!

This addresses lots(!) as it is the changes I tried to send earlier plus more.  There is stuff related to material look, custom margins, option menu handling and good handling of song section trim options.  I appreciate that makes it difficult to go through!

I am hopeful that we have processSong generating good sections and adjusting content as needed - a key process.  A lot of catching up for v6!

If I can get these commits cleared them maybe my Android Studio will be healthier!

I hope they help.

Kind regards
Ian


iv-gha:
For information it does not seem possible to control the padding size to the left and right of the title /clock area of the action bar. They can sometimes be completely removed!  This is the behaviour of the Material theme.  If you change 'Action bar left size with layout bounds displayed you will see it auto adjust the padding as the size changes.  As a compromise I have added 8dp padding on the right of our items to avoid getting too close to the right buttons should the them make it's padding 0.

thebigg73 (approved):
I'll pull these in.  The menu looks like you've replaced the settings cog with the close icon, but I'll check when I pull.
Really hoping this doesn't overwrite my recent updates - there's definitely something not quite right with the pull comparison, but there are no obvious conflicts this time.

As always, thanks for your efforts!!!  Much appreciated

thebigg73:
I see what you did with the close button!  Ignore me!

iv-gha:
Deliberate, yes. It is safest to use the close icon in the xml as generally
that is what is needed and setting it could be missed.

PR #209, opened 2023-03-27
Fix: Consolidated option bar element handling code into navController…

iv-gha:
….addOnDestinationChangedListener - this is, I hope, a reliable place to ensure it is called.

### DIFF
--- a/app/src/main/java/com/garethevans/church/opensongtablet/MainActivity.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/MainActivity.java
@@ -890,17 +890,12 @@ public class MainActivity extends AppCompatActivity implements MainActivityInter
                     displayAreYouSure("exit", exit_confirm, null,
                             Objects.requireNonNull(navController.getCurrentDestination()).getNavigatorName(),
                             navHostFragment, null);
-                } else {
-                    settingsOpen = true;
-                    super.onBackPressed();
                 }
             } catch (Exception e) {
                 e.printStackTrace();
-                super.onBackPressed();
             }
-        } else {
-            super.onBackPressed();
         }
+        super.onBackPressed();
     }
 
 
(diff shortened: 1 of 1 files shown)


thebigg73 (approved):

thebigg73:
Quick testing confirms that this is working without issue now. 
I'm going to have a look at testing the nearby reconnect issue tonight.  Whilst everything isn't fixed or perfect, are you happy for a release of beta 11?  I think there are enough fixes and improvements to warrant it.

iv-gha:
Can you look at the issue of not transposing using flats?  It would be great to sort that.  If only so I can use connect to check transpose  against v5.

Trying not to depress you...   v6 is still playing catch up - I am focused on testing base behaviour and usability and I am aware of lots more to do.  After that there is new features testing.

Stuff should go to beta - the trick is making sure we have time to fix our fixes!  (Smile)

thebigg73:
Fixed the transpose vs user preference option.  It was down to the settings
saving the preference as prefKey_Ab etc, but the transpose looking for
prefKeyAb, etc.

Do we need the capo information in the song title bar in addition to the
capo on screen icon?  Small devices end up using an ellipsize on longer
titles/chords.  The chords can resize, but at the expense of the song title.

thebigg73:
Let me know what else you find as you find it - I'm keen to get things
working correctly.

I must admit that I've been using the beta/alpha versions of v6 live for
well over a year and personally much prefer pretty much every part of it
over v5.  One of the band mates I play with still uses v5.6 and I find it
frustrating when I try to show him different parts of the app and find it
very clunky and much slower.  That said, every user has their own use case
(I use performance mode only with full autoscale, portrait mode, no pads,
no metronome, occasional highlights, inline set).  It's been rock solid for
me.  It's good to have the input of someone using the app in different ways
(e.g. stage mode, etc).

Given the number of fixes made to beta 10, I'm going to push this to beta
11.

Many thanks for your enormous contribution to the project as always!
Gareth

PR #217, opened 2023-04-02
Fix: ProcessSongIntoSections - extra line no longer added by section …

iv-gha:
…name carry forward code.


thebigg73:
This fixes it for Performance mode, but there is a rogue code for Stage Mode

image

iv-gha:
Ah you are so quick...  I had hoped to add fix before you committed this pull!

thebigg73:
Just happened to be tinkering with code - just like you!😉

PR #220, opened 2023-04-03
Fix: ProcessSongIntoSections - double new line split handling

iv-gha:
Hello Gareth,

This avoids lots of empty sections when double new line split processes a run of empty lines .  The code reflects a view that runs start a section (/n/n) that begins with blank lines (more /n).

So

/n/n/n/n/n/n Lyric/n

is processed as '[]/n/n/n/n Lyric/n' which after trim is '[]/n  Lyric/n' - 1 section.

and not process each /n/n as a section which would make it

'[]/n[]/n[]/n Lyric/n' which after trim is  unchanged as '[]/n[]/n[]/n Lyric/n' - 3 sections.

I believe this is better.

### DIFF
--- a/app/src/main/java/com/garethevans/church/opensongtablet/MainActivity.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/MainActivity.java
@@ -947,7 +947,7 @@ public class MainActivity extends AppCompatActivity implements MainActivityInter
             // IV - We are changing so adjust option menu elements
             if (globalMenuItem != null) {
                 // IV - To smooth teardown, we clear elements left to right
-                if (!settingsOpen && whichMode.equals(performance)) {
+                if (settingsOpen && !whichMode.equals(presenter)) {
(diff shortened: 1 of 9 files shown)


iv-gha:
Added some minor moving of section handling code into ProcessSongIntoSections - where it should be!

I have added the missing nearby delayed song send /pending section behavior from v5.  This has separate timing to song display.  The first song change is sent over nearby immediately, further changes of song are delayed 3s with each change resetting the delay.  After settle on a song, 3s later the last song is sent over nearby. (The local song display may have shown a number of songs before the user settles on one, only the last is sent over nearby).  Pending section logic ensures the send is preceded by a pending section request that makes a receiving client open the received song at the current section on the host at the time of the nearby song send.  Ouch!

Whilst I hope I have got it right - we may yet fix the fix!

Cheers
Ian

thebigg73 (approved):
Looks good!  Thank you as always!

PR #222, opened 2023-04-07
Enh: v6 catch up with v5 SongSelect import.  Fix: Transpose preferred keys.

iv-gha:
Hello Gareth,

This upgrades v6 with v5 improved SongSelect extract logic.  Supports extract to xml for both the 'Lyrics' and 'Chords' tab.

There is something missing though.  v5 rightly also calls the click of the SongSelect download button - the downloaded PDF also added to opensongapp.

v6 should do a little more.. the "Chords" SongSelect page needs the download button 'clicked, the  "lyrics" SongSelect page also has a download button that needs clicking.   Also, we should consider support for download on the music sheet tabs as well, these would be PDF sheet music.  For the PDF files it would be good to add, if available, info from the page to the non opensongapp song DB.   We should try to handle the user directly using the SongSelect download button as well.

Following the xml extract the call of the SongSelect download 'click' to get and save the songselect PDF is outstanding in v6.  To keep our users use of SongSelect correct (within term) this needs adding asap.

Happy to discuss.

Cheers
Ian

### DIFF
--- a/app/src/main/java/com/garethevans/church/opensongtablet/chords/Transpose.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/chords/Transpose.java
@@ -191,10 +191,6 @@ public class Transpose {
     private int major;
     private int root;
 
-    private boolean forceFlats;
-    private boolean usesFlats;
-    private boolean capoForceFlats;
-    private boolean capoUsesFlats;
     private String capoKey;
     private int oldChordFormat, newChordFormat;
     private String transposeDirection;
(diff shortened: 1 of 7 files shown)


iv-gha:
I added an account to SongSelect to get the 'free' level of access. On login I ignored the select an organisation bit.   This account could now be used for testing of opensongapp.  I searched for 'Amazing grace' and found it has a public domain version with full lyrics. chords and sheet music - so it is good for testing.

thebigg73 (approved):

PR #228, opened 2023-04-20
Update strings.xml in values-ru

sorrowwind:


thebigg73 (approved):
Looks great and well done!  These will go into beta 16!

PR #229, opened 2023-04-27
Fix: SongSelect extract/download fixes. MYJSInterface setDownload use…

iv-gha:
…d for 'Normal' download (similar to v5) to get download working. Song save logic after direct use of a SongSelect download button improved.

Fix: Updates of NonOpenSong DB now persist - user copy of DB is now written.

Fix: Bad presentation order now results in no presentation order - was causing a blank song with no ability to change the song.


iv-gha:
app/src/main/java/com/garethevans/church/opensongtablet/importsongs/MyJSInterface.java
app/src/main/java/com/garethevans/church/opensongtablet/importsongs/SongSelect.java

These 2 have significant changes that make SongSelect extract work.

The others are, I think,  onResume changes which are small and related to loss of title on leaving and returning to the app.

Doh.

Cheers
Ian

PR #231, opened 2023-05-11
Enhancements and fixes incl. Section processing, Zoom reset

iv-gha:
Hello Gareth,

A section processing fix (another glitch fixed) and other things.  Get them while they do not have commit clashes!

Hopefully good.

Kind regards
Ian


thebigg73 (approved):
Look fine - will pull and test

thebigg73:
All look good Ian - thank you again!

What's your thoughts on how close we are to moving the beta version over to
the stable channel?

iv-gha:
I am not confident of Stage and Presenter modes.  Stage mode has become the hybrid of performance and presenter, it is the way to use a performance songbook and get presenter projection - at this point these modes are not robust and usability in v6 is degraded.

I think work is needed on the settings menu.   I recognise the intent to provide more detail on use but my first (new user ish) reaction was of information overload and a very v5.1 feel with things a little harder to locate.

Google has been considering the tensions inherent in complex menus.  We could adopt the simple 'Android Settings Design Guide' principles.  These were used to arrive at the upcoming Chrome settings menu - a complex menu!  They include some icons as they assess that this is helpful for new users , item notes have a list of 3 main features or the currently set value.  When needed, they group related items under headings and use divider lines.  Detailed notes are made available using a help button.  

v5.6 has less menu confusion than v5.1 (it has fewer sub menus, applies consistent ordering for similar sub menus - compare Song and Set - and uses headings and groupings for long menus.  It has some good points we can consider working into v6.

Rather than make this longer... I think v6 is not yet mature enough for main release.  The desired to unveil new features does not, for me, trump the need to have solid core features and usability.  It must not be worse than v5.6.  It is a complex beast and it will take time!

I personally feel the weight of users relying on OpenSongApp for Church activity - we are a small team, I believe, committed to serving them well.  For me, at this time v6, is not yet robust enough for main release.

Kind regards
Ian

thebigg73:
I'll need to spend more time on stage mode and presenter mode obviously.  I
haven't seen any degraded behaviour on any part of v6 compared to v5.6.0,
which in my opinion is ugly, less usable, slower and frustrating to use.
So many bits didn't work reliably for me in v5.6.0 and even I got lost in
the menus which had been added to and grown and mutated over numerous
iterations. The menus in V6 are more like android settings design (I did
use those) and decided against the icons as they ended up being confusing
with so many similar, but subtly different options that would then require
subtly different icons too.   This partly added to confusion in my
opinion.  I have tried to group settings better and included duplicate
settings where required.

I am definitely mainly a user of performance mode and there are no degraded
parts there that I'm aware of?

Please let me know which parts you think are not robust in stage or
presenter mode and I'll focus my efforts there.  Stage mode was always my
intention to be a hybrid of performance and presenter modes.  I'm unsure of
where v5.6.0 is better in these modes.  I know stage mode is your main
mode, so value your input!

I'm looking for suggestions for things to fix!  The parts I commonly use
are now stable (as far as I know!).

Best wishes,
Gareth

PR #232, opened 2023-05-12
Enh: Proposal for change of top spacing of song menu

iv-gha:
UI design fun!


thebigg73 (approved):

iv-gha:
Hello Gareth,

I had hope that this layout would line up icons? So icons across the top and the settings icon of the main screen - as attached.  Does it do this for your devices? 

Screenshot_20230514-210236-706~2

thebigg73:
Afraid not

iv-gha:
I fudged the offset for the song menu line up with main menu.  The commit
has a manual adjust in the margun calc -  looks like this needs to be
properly calculated - as you understand the margins perhaps you can look at
this so that it becomes good on all devices.

Cheers.

PR #235, opened 2023-05-15
Enh: Alphabet index

iv-gha:
Hello Gareth,

An option for 2 char Alphabet index handling - I think it addresses a perhaps too aggressive layout change when column widths change on moving between 1 and 2 char indexes.  It might be a better user experience  - smoother!!

Kind regards
Ian

### DIFF
--- a/app/src/main/java/com/garethevans/church/opensongtablet/songmenu/SongMenuFragment.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/songmenu/SongMenuFragment.java
@@ -448,13 +448,16 @@ public class SongMenuFragment extends Fragment implements SongListAdapter.Adapte
                 final Map<String, Integer> map = songListAdapter.getAlphaIndex(songsFound);
                 Set<String> setString = map.keySet();
                 List<String> indexList = new ArrayList<>(setString);
-                float tvSize = mainActivityInterface.getPreferences().getMyPreferenceFloat("songMenuAlphaIndexSize", 14.0f);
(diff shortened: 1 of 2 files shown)


thebigg73 (approved):

thebigg73:
Top padding does not align to top od the song scroll area
image

thebigg73:
32dp rather than 40dp seems correct on my device

iv-gha:
Probably needs a rework to have a blank equivalent to song song title above
it.  I spotted on mine that I had not quite aligned - but it seems worse in
your device.

thebigg73:
Feedback on the two initial alphabetical index (from the user who first requested it):

> May I ask why you changed the Two character alphabetical index to vertical instead of horizontal? I find that very very hard to read actually.
> 
> I recommend the following now as I played with the size of the initials:
> 
> Our target is to show as much initials as possible to avoid scrolling or extra clicks and that by having enough space to hit the right inital (pair) with the first try while not clicking on the song occasionally, right?
> Currently the size setting is for the Font what I find suboptimal. My main intention for it when  I asked for it was for a wider column.

iv-gha:
No problem,  it is good to get UI feedback.

The aim was to prevent an unsightly layout change as the column width
changed - this makes the app somewhat ill behaved in transition.

You will see that the space between rows is less when showing 2 chars than
when showing 1 - this is what the code happened to do and it seemed good to
me.  1 or 2 char are easy to click (our click area considerations).

An alternative is a fixed width 2 char column - open to the repost of 'why
waste space'.

I default to choosing layouts that give the app a professional look so
would not like reversion.

It is change and that can be ill received.  I suspect a version changes
gets quite a few views expressed!

PR #243, opened 2023-06-21
Update RU strings.xml

sorrowwind:
Some newly added strings translated.


thebigg73 (approved):
Many thanks for these!

PR #244, opened 2023-06-22
Enh: Second screen work

iv-gha:
Hello Gareth,

Work on Second Screen render behaviour...  Info bar plays nicely with logo, blank, black features now.  Cross fades are a little smoother.

The View alpha panic has been removed as it was interfering - I think the use of 2 views for content and info is a great improvement and probably means a panic is not needed.  Continuing to test.  If a panic proves to be needed then it perhaps needs to be smarter.

Hopefully good
Ian


app/src/main/java/com/garethevans/church/opensongtablet/customviews/SongProjectionInfo.java
```diff
@@ -170,6 +178,13 @@ public void setupFonts(MainActivityInterface mainActivityInterface) {
             songCCLI.setTypeface(typeface);
             textClock.setTypeface(typeface);
 
+            // IV - A dialog uses white text
```
thebigg73, line 181: Just checking - aren't the colours defined by the theme (presoInfo font colour in Presentation mode, or lyric font colour in Performance/Stage modes?)

iv-gha:
The white is for the song info at the top of presenter mode screen only.

Also in this case it is set for single lines.  Other cases use the theme color and multi lines.

Cheers

thebigg73:
Ah, that make sense - previewing code in PRs doesn't always give the full picture!

PR #247, opened 2023-07-07
Enh: SongSelect extract

iv-gha:
Hello Gareth,

Lots of changes to support new format used for lyrics only download and some further refinement of the excellent flatten of shadow Dom approach.  All the SongSelect download buttons, direct use on the web page, do sensible things (I hope).

Take a look!

Thanks
Ian


thebigg73 (commented):
Looks good Ian!
I'll pull in and check - I see you also spotted and fixed my 3s delay was being applied to everything and this was why I thought I'd broken the other online imports!
The 3s was arbitrary, but seemed to be enough for most cases.  Clicking on the search icon will do the check again anyway, so there is a way around it if not - I didn't want to add in a secondary check for the sake of it.

PR #248, opened 2023-07-10
Enh: SongSelect

iv-gha:
Hello Gareth,

I've tried different JavaScript methods to wait for the Shadow DOM expansion as opposed to a fixed time, but they haven't been effective. It seems the webview isn't updating when we manipulate it.  I think this is because Webview is not Shadow DOM aware.

I've made some adjustments so that the delay only occurs on pages with viewers, detected by the site URL. This detection remains consistent even when changing site country settings. I've set a 5-second delay to ensure the view fully builds before extracting the information. This way, the delay only applies when needed, not for general navigation on SongSelect.

A few other tweaks too.

Kind regards
Ian


thebigg73 (approved):
Many thanks as always!

PR #252, opened 2023-12-07
PL translations update

mjdominiczak:
Hello Gareth,

I've been using OpenSong in Poland for some time already and I thought that contributing here is a must. I am both musician and Android dev myself and I just wanted to say I am impressed with the amount of functionality you've managed to put into the app. The code is also of great quality and it is really nice to explore, although there's none of fancy Kotlin or Compose stuff :)

I would like to adjust Polish translation of the app - I've done my best to grab the correct context. Would be great to see it in the next release.

That being said, I cannot build the app from the most recent commit. There are some string keys missing - it would be nice to fix it and translate before merging too. Can you have a look? Missing keys are used in `settings_pedal.xml` file: 

```
string/pedal_test_info
string/pedal_test
string/pedal_detected
string/repeat_mode_info
string/repeat_mode
string/repeat_mode_time
string/repeat_mode_count
```

Best regards,
Michał Dominiczak


thebigg73:
I'll pull in your current translation updates and add in the remaining translations - these aren't currently used in the app but are something I'm testing out.

thebigg73 (approved):
Looks great!  Thank you for taking the time to do this!

PR #262, opened 2024-05-22
Fix for capitalized letters in font resource name

LudovicAL:
Hi Gareth,
English is my second language.
In attempting to deploy and run your app for the first time, from Android Studio, to my physical tablet in debug mode, I encountered the following error:

> java.lang.RuntimeException font asset not found font/robotomono.ttf

The error occured while executing line 111 in class _app/src/main/java/com/garethevans/church/opensongtablet/appdata/MyFonts.java_.

Please find in this Pull Request a fix that solves the error.

Strangely enough, the error does not occurs when the app is installed on the same device from the Google Play Store.
I am assuming the error is not happening for you, or for most contributors.
The fix could nevertheless spare those in my situation from a minor headache.


thebigg73:
Thank you for this!
Something weird has happened.  On my device the Roboto font file is named robotomono.ttf, but the GitHub file is named RobotoMono.ttf.  This is why my version and the Google Play version is fine, but your fork isn't.  I'll update the font file too.

PR #271, opened 2025-11-07
Fix web server to bind to WiFi network instead of mobile data

dhsutton:
When both WiFi and mobile data are active, the web server now correctly prioritizes the WiFi network interface, ensuring nearby devices on the local WiFi can connect. Previously, it could bind to the mobile network's carrier-grade NAT address (100.x.x.x), making it unreachable.

Changes:
- Prioritize WifiManager API for most reliable WiFi IP detection
- Filter out mobile network interfaces (rmnet, ccmni, pdp)
- Filter out carrier-grade NAT addresses (100.64.0.0/10)
- Explicitly prioritize wlan/eth interfaces over others


thebigg73 (approved):
Done

PR #272, opened 2025-11-07
Fix background images/videos to use relative paths for cross-device sync

dhsutton:
When background images or videos are stored in the OpenSong folder, they are now saved with relative paths (../OpenSong/...) instead of absolute device-specific paths. This allows settings to sync correctly across different devices (phones, tablets, Chromebooks, etc.) where the absolute path structure differs.

Implementation:
- Updated ImageChooserBottomSheet.java to use fixUriToLocal() when saving background URIs
- Applied to backgroundImage1, backgroundImage2, backgroundVideo1, and backgroundVideo2
- Files inside OpenSong folder use relative paths (portable)
- Files outside OpenSong folder use absolute paths (device-specific)
- Maintains backward compatibility with existing absolute paths
- Uses existing infrastructure already proven for custom logos

Technical Details:
- ImageChooserBottomSheet.java lines 270, 277, 284, 291
- Changed from uri.toString() to fixUriToLocal(uri)
- fixUriToLocal() detects "OpenSong/" in path and converts to relative
- fixLocalisedUri() on load converts relative paths back to absolute URIs


thebigg73:
Looks good - many thanks for submitting these!

PR #275, opened 2025-11-12
Fix: Add defensive case-insensitive font loading

dhsutton:
Problem:
- Android filesystems are case-sensitive but development environments (Windows/macOS) often aren't
- Hard-coded font paths like "font/Lato.ttf" crash if actual file is "font/lato.ttf"
- App crashed when entering Presenter Mode due to font loading failure

Solution:
- Added loadFontSafely() method that handles case mismatches:
  1. Tries original path first (fast path)
  2. Lists directory and finds case-insensitive match
  3. Falls back to system SANS_SERIF font if not found
- Updated all 4 font loading locations in MyFonts.java

Benefits:
- App never crashes due to font filename case mismatches
- Works correctly across all platforms (Windows/macOS/Linux/Android)
- Handles any case variation automatically
- Graceful degradation with clear logging
- More maintainable than trying multiple case variations

Tested:
- Normal app launch
- Presenter Mode (previously crashed)
- Devices without Google Play Services
- Google Font download failures

Fixes the crash reported when switching to Presenter Mode.

### DIFF
--- a/app/src/main/java/com/garethevans/church/opensongtablet/appdata/MyFonts.java
+++ b/app/src/main/java/com/garethevans/church/opensongtablet/appdata/MyFonts.java
@@ -108,7 +108,7 @@ public class MyFonts {
         changeFont("fontPreso",fontPreso,presoFontHandler);
         changeFont("fontPresoInfo",fontPresoInfo,presoInfoFontHandler);
         changeFont("fontSticky",fontSticky,stickyFontHandler);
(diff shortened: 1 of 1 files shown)


thebigg73 (approved):
The file in the assets folder (Lato.ttf) does start with a capital, so I will adjust this as the fallback (font/lato.ttf > font/Lato.ttf).
I have also bundled a different font for monospace that I personally prefer (Inconsolata.ttf), so will also adjust the RobotoMono.ttf to this.

dhsutton:
Sounds good! Thank you.

